PR #4, opened 2018-02-05
Job edits

ChristinaLK:


_episodes/12-scheduler.md
```diff
@@ -165,12 +170,18 @@ which is probably not what we want.
 
 The following are several key resource requests:
 
+* `-n <nnodes>` - how many nodes does your job need?
```
jstaf: I'll admit I didn't want to get too far into multi-node jobs for these intro materials... that said, `-n` specifies number of tasks which does not necessarily correspond to number of nodes if `-c` is set (cpus per task). Could you add a bit of clarification here on this and then I'll merge this pull request?

ChristinaLK: I was lazy and didn't look up the right option for nodes ( `-N`).  Oops. 

I was thinking bigger picture -- having a list of the common options that people should know about (whether or not they need them immediately).

jstaf: Ok sounds good - I will add some clarification and merge this in (I might move the an overview of the differences between `-n` and `-N` into a callout box somewhere).

jstaf:
In light of the potential upcoming changes to this repository, I'm going to merge this for now and add my original `-n`/`-N` distinction as an issue on the repository. (trying to get the remaining pull requests merged)

PR #32, opened 2018-08-28
spelling corrections using en_GB dictionary

tkphd:
Ran `ispell` on each markdown file in `_episodes` using the `en_GB` dictionary, based on the British spellings in the first episode. Numerous spelling mistakes are corrected.


psteinb:
as mentioned in hpc-intro, we need to decide on a dictionary to use first. Sorry, but I don't feel to be in the position to make this call alone ... both from a community and language perspective.

psteinb:
according to carpentries/lesson-example#231 en_GB should be the dictionary to use, so I'd strongly vote to not merge en_CA.

PR #52, opened 2018-09-19
More clarity about user state and cluster heterogeneity

### LINKED ISSUE #45: Elaborate on "Efficiency" in "Why use a cluster?"
The "Why Use These Computers?" section of this lesson has a discussion of "efficiency", which mentions large clusters are often a "pool of resources drawn on by many users" and mentions that they can be in use constantly. There's another side to this coin, which is that projects or groups may have a requirement for very powerful computers, but only part of the time, so sharing the resource makes economic sense for them also.
This dovetails a little bit with the existing "what is a cluster" and "definition of cloud" conversations -- pay-per-use is a much-remarked-on feature of the cloud, and relevant to institutional HPC also.

### LINKED ISSUE #46: "Working on a cluster" -- assumption of user log-in state
The "Working on a cluster" episode begins with the assumption that the user is logged in to the head node of a cluster ("what computer have we logged into?"), but the actual step of logging in to the head node is not called out, either at the start of this lesson, or the end of the previous one. Since the point of the description is to clarify that there are multiple contexts or scopes within a cluster, it's important to make this step explicit.

### LINKED ISSUE #47: "Working on a cluster" -- nodes may be heterogeneous
In the "Working on a cluster" episode, where contexts and scopes within the cluster are introduced, it's worth pointing out that the cluster's node collection may be heterogeneous. It's probably overkill to get into resource requests, but a remark that nodes are not always all equal would be valuable here.

### LINKED ISSUE #51: "Using resources effectively", "Measuring stats" -- bad description of SSH/SLURM relationship
In the "Using resources effectively" episode, under "Measuring the staticstics of currently-running tasks", there's text that says "One very useful feature of SLURM is the ability to SSH to a node where a job is running...". 
The SSH-ability of nodes is a result of cluster configuration independently of the queuing system. SLURM provides interactive access, but direct SSH may be disallowed even in the presence of SLURM, or allowed for other queuing systems.

Propose to reword this to "Typically, clusters allow users to SSH directly into worker nodes from the head node. This is useful to check on a running job and see how it's doing."

reid-a:
Several minor edits for additional clarity about the learner's state and system ops.
Closes #45, closes #46, closes #47, closes #51.

### DIFF
--- a/_episodes/11-hpc-intro.md
+++ b/_episodes/11-hpc-intro.md
@@ -56,7 +56,7 @@ Using a cluster often has the following advantages for researchers:
   projects.
 * **Efficiency.** Many HPC systems operate a pool of resources that are drawn on by many users. In
   most cases when the pool is large and diverse enough the resources on the system are used almost
-  constantly.
+  constantly, with individual users only using (and paying for) a fraction of the system.
 * **Cost.** Bulk purchasing and government funding mean that the cost to the research community for
   using these systems is significantly less than it would be otherwise.
 * **Convenience.** Maybe your calculations just take a long time to run or are otherwise
--- a/_episodes/12-cluster.md
+++ b/_episodes/12-cluster.md
@@ -35,6 +35,8 @@ too big for any one computer.
 
 ## Where are we?
 
+Go ahead and log in to the cluster, using the `ssh` command.
+
 Very often, many users are tempted to think of a high-performance computing installation as one
 giant, magical machine. Sometimes, people will assume that the computer they've logged onto is the
 entire computing cluster. So what's really happening? What computer have we logged on to? The name
@@ -148,6 +150,10 @@ the computer's memory. Disk is a computer's long-term storage for information it
 > ```
 > {: .bash}
 {: .challenge}
+
+> ## Differences Between Nodes
+> The cluster may not be homogeneous. Many HPC clusters have a variety of nodes optimized for particular workloads. Some nodes may have larger amount of memory, or specialized resources such as Graphical Processing Units.
+{: .callout}
 
 > ## Units
 > 
--- a/_episodes/16-resources.md
+++ b/_episodes/16-resources.md
@@ -97,8 +97,7 @@ Some interesting fields include the following:
 
 ## Measuring the statistics of currently running tasks
 
-One very useful feature of SLURM is the ability to SSH to a node where a job is running and check
-how it's doing. To do this, check where a job is running with `squeue`, then run `ssh nodename`.
+Typically, clusters allow users to SSH directly into worker nodes from the head node. This is useful to check on a running job and see how it's doing. To do this, check where a job is running with `squeue`, then run `ssh nodename`.
 
 However, we can also check on stuff running on the login node right now the same way (so it's not
 necessary to `ssh` to a node for this example).


_episodes/11-hpc-intro.md
```diff
@@ -56,7 +56,7 @@ Using a cluster often has the following advantages for researchers:
   projects.
 * **Efficiency.** Many HPC systems operate a pool of resources that are drawn on by many users. In
   most cases when the pool is large and diverse enough the resources on the system are used almost
-  constantly.
+  constantly, with individual users only using (and paying for) a fraction of the system.
```
psteinb: I'd indicate that they are "possibly" paying for it

_episodes/12-cluster.md
```diff
@@ -35,6 +35,8 @@ too big for any one computer.
 
 ## Where are we?
 
+Go ahead and log in to the cluster, using the `ssh` command.
```
psteinb: Please add a code block that does ssh. Let's hard code the machine name as `remote` for now.

_episodes/12-cluster.md
```diff
@@ -149,6 +151,10 @@ the computer's memory. Disk is a computer's long-term storage for information it
 > {: .bash}
 {: .challenge}
 
+> ## Differences Between Nodes
+> The cluster may not be homogeneous. Many HPC clusters have a variety of nodes optimized for particular workloads. Some nodes may have larger amount of memory, or specialized resources such as Graphical Processing Units.
```
psteinb: I suggest to expand on homogeneous here. For example adding "i.e. all compute nodes consist of the same model by the manufacturer". We could leave it at this or mention RAM, CPU count, etc. ... which in turn would force us to define all of those first. Here is a great chance to add conceptual pictures, schemas or art here!

_episodes/16-resources.md
```diff
@@ -97,8 +97,7 @@ Some interesting fields include the following:
 
 ## Measuring the statistics of currently running tasks
 
-One very useful feature of SLURM is the ability to SSH to a node where a job is running and check
-how it's doing. To do this, check where a job is running with `squeue`, then run `ssh nodename`.
+Typically, clusters allow users to SSH directly into worker nodes from the head node. This is useful to check on a running job and see how it's doing. To do this, check where a job is running with `squeue`, then run `ssh nodename`.
```
psteinb: let's not use `ssh` as a verb, that is community lingo. Did we introduce compute nodes or worker nodes before? I am afraid, the concept of a head node does not exist in the cloud, right?

_episodes/16-resources.md
```diff
@@ -97,8 +97,7 @@ Some interesting fields include the following:
...
```
psteinb: should we really encourage people to ssh into nodes?

_episodes/11-hpc-intro.md
reid-a: No objection.

_episodes/12-cluster.md
reid-a: No objection, can do.

_episodes/12-cluster.md
reid-a: Possibly the best solution is to just remove the first sentence, so that it just reads "Many HPC clusters have a variety ..."
If that's too terse, we could invite the learners to compare the specs on "their" node to others.

_episodes/16-resources.md
reid-a: Suggest " ... clusters allow users to connect directly to worker nodes".  The fact that this operation is via ssh is implicit in the command, which is later in the same sentence.

_episodes/16-resources.md
reid-a: Maybe not "encourage", but it can be a useful diagnostic tool, it seems reasonable to me that learners should be aware of it. We could add some cautionary text about only doing it for diagnostic purposes, and not as a way of cheating the scheduler.

_episodes/16-resources.md
psteinb: sounds good to me.

_episodes/12-cluster.md
psteinb: good idea - go for it. what I typically do in my workshops is having the participants compare the specs of a node with those of their laptop. this may take time as finding the processor type etc in different OSes can be challenging. But I feel that its a worthwhile task as they can relate better to what they are working with. In the future, we might even have them compare with the specs of their mobile phone or so.

_episodes/16-resources.md
psteinb: that's exactly my concern. let's pack it into a call-out for now as a compromise. In the end, all users need to have ssh access to all compute nodes as e.g. MPI uses that to administer the processes.

_episodes/12-cluster.md
reid-a: I have added a code block, but I unresolved this because you might want another look (once I've pushed the change) -- I've provided a verbose [user@laptop]$ prompt to establish scope, but I can see where this might be extraneous detail.

_episodes/16-resources.md
reid-a: I've put in a call-out about SSH, but the coordination with the following content is now fairly poor, that section talks about running "top" and "free" to see what's going on, but this only makes sense on the node, which we are now advising against?
This needs more attention.

Issue #48, opened 2018-09-18
"Scheduling jobs" -- remark about environment variables is incorrect


reid-a:
In the "Scheduling jobs" episode, under the "other types of jobs" subsection, there is a remark that environment variables are not available for interactive jobs launched via "srun". On the two SLURM clusters available to me, this is not true for either of them -- doing "srun" followed by "env" shows that in the shell session, the "SLURM_*" environment variables are set, and the explicitly-mentioned SLURM_CPUS_PER_TASK variable is set if the "-c" flag was provided to "srun".

This might be configurable, and vary from one installation to another?

psteinb:
Can you pinpoint me to the specific position in the episode, please?

reid-a:
It's towards the end of the episode. In the "Other types of jobs" sub-heading, there's a paragraph that begins "srun takes the same options as sbatch" -- the parenthetical remark at the end of that paragraph is the wording I'm referring to.

PR #53, opened 2018-09-28
Reworded the text about SLURM environment variables.  Fixes issue #48.

reid-a:
The new wording is a bit weasely, it says "typically" srun environments are like sbatch ones. Open to additional discussion.


psteinb:
thanks.

PR #54, opened 2018-10-11
Fixed detailed memory command

jdhayes:


tkphd:
@psteinb the PR fixes a typographical error. In the current [12-cluster.md](https://github.com/hpc-carpentry/hpc-intro/blob/gh-pages/_episodes/12-cluster.md), the line reads
```
cat /proc/meminfo free -m
```
This PR corrects that to the intended
```
cat /proc/meminfo
```

Thanks for the patch, @jdhayes

tkphd:
`free` is still the preferred/recommended command. `/proc/meminfo` is mentioned "for more details." I wouldn't use it in a live lesson, since the output is lengthy, but it's still useful to show an alternative way to access hardware information.

For a full accounting, look at the [blame log](https://github.com/hpc-carpentry/hpc-intro/blame/gh-pages/_episodes/12-cluster.md#L140). @ChristinaLK may want to weigh in on fixing the typo *vs.* eliminating the command.

jdhayes:
The free command is still used in the code segment immediately above this fix.
I guess I assumed since the free command was mentioned twice, and the second time it was mentioned the code does not run as is, that this was a typographical error.
This patch would only allow the detailed and the free methods to both work as intended, there is no loss of information.

Issue #63, opened 2018-11-15
login node limits


baberlevi:
17-responsibility says: 'A “quick test” is generally anything that uses less than 10GB of memory, 4 CPUs, and 15 minutes of time. Remember, the login node is to be shared with other users.'

This is going to vary widely by site. Many will have enforced ulimit or cgroup limits on the headnode that are significantly lower than this and may cause this type of 'quick test' to fail when it runs out of memory. 

How about just, don't run on the login node, and explain about looking for a debug or interactive queue ?

psteinb:
Good idea, please submit a PR.

PR #66, opened 2018-12-07
login limts

baberlevi:
Addresses issue #63


baberlevi:
I added the liquid var and tested, thanks for the suggestion

psteinb:
thanks for you contribution. That is a wonderful use of liquid there.

PR #68, opened 2018-12-12
Attempt to fix base-url on deployed site.

Carreau:
As afar as I understand doing so should prepend the baseURL to the
figure thus making the figure correctly appear in gh pages.

It seem to work locally when forcing the baseurl with the --baseurl
flag.

Also when URL for figure is not set, do not inject the a tag which
create an empty link.


psteinb:
It's hard for me to test this.

PR #71, opened 2018-12-13
added site.url in liquid statements to make the images display on git…

### LINKED ISSUE #67: setup jekyll baseurl/relative_url
most of the image are broken as they try to load from `/fig/` instead of `/hpc-intro/fig/...` this seem to be due to the missing `baseurl` config. 

It [seem like](https://jekyllrb.com/news/2016/10/06/jekyll-3-3-is-here/#2-relative_url-and-absolute_url-filters) you can use filters to prepends only on github, and not when deploying/testing locally.

psteinb:
…hub.io; trying to fix #67

### DIFF
--- a/_episodes/11-hpc-intro.md
+++ b/_episodes/11-hpc-intro.md
@@ -56,7 +56,7 @@ Let's dissect what resources programs running on a laptop require:
 
 Schematically, this can be reduced to the following:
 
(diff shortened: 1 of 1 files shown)


psteinb:
just saw that this is a dublicate of #68

PR #74, opened 2018-12-17
Fix SBATCH option names and effects.

### LINKED ISSUE #72: Mistake in sbatch options ?
```
Resource requests
But what [.....], which is probably not what we want.

The following are several key resource requests:

-n <nnodes> - how many nodes does your job need?

-c <ncpus> - How many CPUs does your job need?
```

- I believe `-n` is the number of tasks, the number of nodes is uppercase `-N`, 
- `-c` is the number of CPU but per task, written as above, it feels like it's the total number of CPU.

Carreau:
Closes #72.

I did not find any good way to not introduce node/task/cpu.
Especially the task/cpu on single nodes an people not familiar with
threading and multiprocessing can be confusing so i'm unsure.

### DIFF
--- a/_episodes/13-scheduler.md
+++ b/_episodes/13-scheduler.md
@@ -162,9 +162,11 @@ site's default resources, which is probably not what we want.
 
 The following are several key resource requests:
 
-* `-n <nnodes>` - how many nodes does your job need? 
+* `-N <nnodes>` - how many nodes does your job need? 
 
-* `-c <ncpus>` - How many CPUs does your job need?
+* `-n <ntasks>` – how many tasks does your job have?
+
+* `-c <ncpus>` - How many CPUs per task does your job need?
 
 * `--mem=<megabytes>` - How much memory on a node does your job need in megabytes? You can also
   specify gigabytes using by adding a little "g" afterwards (example: `--mem=5g`)


Carreau:
I think that make sens. 

I believe if we just use `-c` then you are guarantied to have `n=1` and `N=1`, we can add a sentence that say something like "In the tutorial we'll restrict ourself to computation that do not scale beyond a single machine", or "Running computation across multiple machine is beyond the scope of this introduction, and we invite you to consult other resources or the advance tutorial".

psteinb:
Yes, that sounds like a superb idea to move forward.

tkphd:
Maybe we can move this content into Extras?

CarrieBrown:
I would almost like to see this use the long flags (such as `--nodes` as opposed to `-N`) since this can be more intuitive for novices (since shorthand flags can cause confusion such as `-N` vs `-n`).

psteinb:
@Carreau If you find the time resources, please update this PR minding @CarrieBrown 's comments on long flags. Then I think we are getting closer to a merge. If you need help, please let us know.

tkphd:
Just submitted a PR against @Carreau's fork to rebase this contribution and implement @CarrieBrown's revisions: [carreau/hpc-intro #1](https://github.com/Carreau/hpc-intro/pull/1).

Issue #6, opened 2018-02-05
Potential images or diagrams


ChristinaLK:
* screenshot of putty for the ssh episode
* diagram of two computers, for ssh episde
* diagram of cluster (head node + worker nodes) for cluster episode

kamil963:
I have given the following intro to scheduling talk, 
there are many potential diagrams in that presentation to simplify and illustrate 
concepts. I can edit them as well for specific example for addition to HPC carpentry.

https://docs.google.com/presentation/d/e/2PACX-1vQFH3oQL6NAOogswftWy19E1jwIkzO0lFzNKQVXKRPX4QaxCh3SB4EYxg3b7QlXCNEP7k6x6j8DYHDh/pub?start=false&loop=false&delayms=3000

Sabryr:
Scheduler  - in our current training material we depict scheduler as a "bouncer"  manger a queue for crowded club (Slide 17 of https://www.uio.no/english/services/it/research/events/2018b/abel_intro_march2018.pdf) . If this makes sense, I can create  a diagram (we do not have a citation for the current diagram) with CC-BY.

ChristinaLK:
I love it! I've definitely compared a scheduler to the host/hostess at a restaurant, which is the same idea.

Sabryr:
OK, I take that positive response as an encouragement and make a SVG (easier to modify and version control friendly) . I think the restaurant idea is more closer as we have tables with fixed number of seats.  When you meant  host/hostess I guess you were thinking more of when you arrive at the door and then someone take you in when a table is empty .

ChristinaLK:
@Sabryr yes, that's what I meant.  I also really like that analogy because (at least on our systems), jobs that request fewer resources will start sooner, just like smaller parties get seated faster at a busy restaurant. ;)

ocaisa:
@bernhold I agree about the footnote, our site is the same.

tkphd:
@Sabryr and @ChristinaLK, that'd be the *maître d'* leading you to an appropriately sized table, once one becomes available.

@bernhold, I think the analogy holds: your facility would be like a restaurant with several very large tables, and few small ones. The medium-sized jobs just have to wait until a suitable table opens up, or until the *maître d'* can find a complementary group to add so that the composite fulls a large table.

tkphd:
*Cross-posted from #84*

[...]

PR #84, opened 2019-02-11
Diagram restaurant queue manager issue #6

Sabryr:
Related to https://github.com/hpc-carpentry/hpc-intro/issues/6

Comments welcome.


Sabryr:
This is my initial suggestion, please contribute to improve this.

ChristinaLK:
I like it. I don't know what I would change except adding more labels like jobs, computers, cpus, etc.

Sabryr:
Not much comments so doing some more refining

ChristinaLK:
I like the concept. The colors are a little dark -- is it easy to change to them to be lighter? I think that would make the diagram easier to read.

Sabryr:
Yes of course colors can be changed (its svg), I will do this as soon as I get time

fig/restaurant_queue_manager.svg
```diff
@@ -0,0 +1,2516 @@
...
+             style="fill:#ffffff;stroke-width:0.26458332"
+             id="path953"
+             inkscape:connector-curvature="0" />
+          <path
+             d="m 79.574019,139.48062 c 0,0 -0.01738,2.05046 -0.01738,2.05046 0,0 -4.335899,3.62075 -4.335899,3.62075 0.0155,2.8815 0.192377,4.94156 0.530304,6.17821 0.341068,1.21677 0.82565,1.55594 1.452754,1.01985 0.362437,-0.30984 0.714061,-0.8075 1.054904,
```
tkphd: Bussy -> Busy

tkphd:
At a glance, the restaurant SVG seems to provide a good analogy for the workload manager / queueing system. On closer inspection, the restaurant suffers from diversity problems.

- Genders are delineated by pants *vs.* skirts, which is a dated visual cue.
- While a few beskirted individuals are enqueued, they are apparently not allowed in the dining room?

I would prefer a major revision to simplify the visual style and strip it of this implicit social commentary. For example, a top-down view with simple geometric shapes indicating a filled or occupied status by coloring would serve an equivalent purpose, reduce the maintenance overhead, and withhold judgment on what sorts of jobs are allowed through the doors.

tkphd:
The metaphor seems to break down the further it stretches. In a restaurant, raw material is converted to finished results by the back-of-house staff, usually hidden in the kitchen: this is the parallel workforce. The front-of-house staff carry the results from the workers to the clients, more like an interconnect or intranet linking the HPC facility to the campus or Internet.

Perhaps better analogies could be drawn between a shared office space, where the workers are the professionals occupying each office. Reservations and access are managed through the front desk (workload manager). Different offices serve different purposes (architectures/accelerators): accounting jobs go to the accountant, legal to the lawyer, *et cetera*. A conference room (interconnect) permits efficient collaboration by temporary associations (communicators) of different professionals (nodes).
A linear workflow can be crafted...

- A client comes through the door, carrying with them their notes and reference materials (SSH into the login node).
- The client requests an appointment at the front desk (submit the job to the workload manager).
- The front desk staff reads through the client's portfolio to assess the workload requirements.
- If none have been specified, or the request exceeds this office's resources, then the job is rejected.
- If the job is manageable, the front desk determines the earliest available time and sets the appointment (returns the job ID).
- Appointments are best-estimates. The actual start time may drift relative to the reported "start time."
- If the task requires only one professional, then at the appointed time, the front desk hands off the client's portfolio and time limit and the worker gets to work.
- Once finished, or out of time, the front desk retrieves the updated portfolio.
- If the task requires more than one professional, then at the appointed time, the front desk calls the necessary workers into the conference room and delivers the portfolio and time limit. The pool of workers get to work, communicating as necessary, until the task is complete or the clock runs out. 
- Once finished, or out of time, the front desk retrieves the updated portfolio.
- The client may ask at the front desk (check job status), or the front desk may contact the client when the job changes state (start, finish, error).
- Once the job is complete, the client may retrieve the portfolio from the front desk (SSH into or RSYNC from the login node).

All that being said, explaining this extended metaphor in detail would be tantamount to describing the real HPC system in detail. I doubt this abstraction helps the learner to understand; it would take a couple walk-throughs in the class to get the facts straight; and it doesn't help anyone actually understand and use an HPC resource. The time would be better spent, in my opinion, in describing increasingly complex computational frameworks:

1. You launch a program on the computer at your desk.
2. You ask a colleague to run the program on their beefier computer. They let you know when it finishes.
3. You modify the program to use all available cores on your colleague's computer.
4. ...
8. You submit your job to the queuing system, and it runs in no time on the HPC resource.

Sabryr:
Thank you for the detailed review. Intention was not to have a one-to-one mapping with a HPC system, rather introduce the concepts to a new comer. IMO, to explain all the concepts would make the diagram more complected than the concepts we are trying to introduce (in-addition to what is discussed we have GPUs, software modules, MPI, QoS,faire-share, quotas, node failure, rescheduling ...  ...).  I will try to refine the image to get closer the review and lets see whether this is feasible to proceed.

tkphd:
I agree that a one-to-one mapping of *every* concept would be overkill. However, unless there's a one-to-one mapping of *key* concepts, the metaphor is just a distraction. My opinion is that, in the case of a restaurant metaphor for an HPC service, there is not enough similarity to make a helpful comparison. If the only point we're trying to make is that the workload manager in an HPC environment is like the *maître d′* at a restaurant, then we can say that in a single sentence without any further illustration and save the time required to describe the restaurant metaphor for other purposes.

ChristinaLK:
I think the key concepts of "work", "resources" and "scheduler" are all usefully conveyed in this metaphor and a diagram helps give those concepts a shape.

tkphd:
- What is the "work," and how is it accomplished?
- What are the "resources," and how are they produced and consumed?

My opinion is that the "scheduler" is the only unambiguous concept in this metaphor.

Sabryr:
Reopening

Sabryr:
Reported issues:
The colors are a little dark : Now lighter, neutral  
Genders are delineated by pants vs. skirts, which is a dated visual cue  :- all are robots now
 While a few beskirted individuals are enqueued, they are apparently not allowed in the dining room :- all are robots now
So the attempts is to convey "Diagram restaurant queue manager ", not to explain HPC.

Sabryr:
@tkphd , do you think it is OK to use this diagram to explain the queue manager ?. sorry that I could not address all your concerns as my attempt was not to explain HPC in its entirety. But I will send PRs for diagrams for parts of the HPC using the valuable input you gave (e.g. the HPC kitchen)

fig/restaurant_queue_manager.svg
```diff
@@ -0,0 +1,6970 @@
...
+    <radialGradient
+       inkscape:collect="always"
+       xlink:href="#linearGradient16487"
+       id="radialGradient16489-3-7-8-0-5-9-2-0-8-5-9"
+       cx="-37.218052"
+       cy="55.575729"
+       fx="-37.218052"
+       fy="55.575729"
+       r="3.2921085"
+       gradientTransform="matr
```
tkphd: Busy

tkphd: Reserved

Sabryr: Thank you and done.

Sabryr: Thank you and done

Sabryr:
This is how it is rendered on my folk:
https://sabryr.github.io/hpc-intro/13-scheduler/index.html

PR #102, opened 2019-09-20
Update 13-scheduler.md

mpiercy:
Should be capital N for nodes -N
-n is number of tasks

-n, --ntasks=<number>
              sbatch does not launch tasks, it requests an allocation of resources and submits a batch script. This option advises the Slurm  controller  that  job  steps  run
              within  the  allocation  will  launch  a  maximum  of  number tasks and to provide for sufficient resources.  The default is one task per node, but note that the
              --cpus-per-task option will change this default.

### DIFF
--- a/_episodes/13-scheduler.md
+++ b/_episodes/13-scheduler.md
@@ -174,7 +174,7 @@ stuck with your site's default resources, which is probably not what we want.
 
 The following are several key resource requests:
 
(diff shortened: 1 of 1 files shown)


mpiercy:
Very true!  Most of our users never use a whole node, its mostly 1 core and 4 GB RAM for a few hours, lots a highly parallel workloads run.

PR #124, opened 2020-06-24
Lesson 4 (accessing software)

mattgillucl:


mattgillucl:
Apologies. This PR was meant for our fork of this repo, rather than this one. Please ignore it.

psteinb:
I'd be curious to learn more about this fork.

dpshelio:
We are converting the bits for University College London's clusters. We will open a PR with that here once it's done. You'll see us also opening some other separate PRs if we find errors or we'd like to propose some changes.
The problem is that when working in a fork, gh by default shows the PR with upstream, and that gets a bit confusing as then the PRs is not happening where you want to happen. :wink:

PR #125, opened 2020-06-24
👷 allows "namd" as a word so the spelling doesn't fail

dpshelio:
It may have been a recent update on `codespell` but now `namd` (a library mentioned in some places) is making the CI to fail. This list will probably grow as people add more examples for different HPC, so I don't know when the codespell will become useless by allowing many libraries.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -23,5 +23,5 @@ jobs:
       # Use codespell for spell checking
       install:
         pip install codespell
-      script: codespell --skip="assets,*.svg,bin" --quiet-level=2  --ignore-words-list="rouge,dropse" -L "namd"
+      script: codespell --skip="assets,*.svg,bin" --quiet-level=2  -L "rouge,dropse,namd"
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -23,5 +23,5 @@ jobs:
       # Use codespell for spell checking
       install:
         pip install codespell
-      script: codespell --skip="assets,*.svg,bin" --quiet-level=2  --ignore-words-list="rouge,dropse"
+      script: codespell --skip="assets,*.svg,bin" --quiet-level=2  --ignore-words-list="rouge,dropse" -L "namd"
```
ocaisa, line 26: Looks like `ignore-words-list` option has been replaced by `-L`, can you please update this line to reflect that:
```
script: codespell --skip="assets,*.svg,bin" --quiet-level=2  -L "namd","rouge","dropse"
```
(Either they are accepting the original option in deprecated mode, or the words no longer kick an error)

PR #126, opened 2020-06-30
Update lesson 13 to be compatible with Sun Grid Engine

tkoskela:
* Add a line ‘This scripts has finished successfully’ to example scripts to be able to tell when it didn’t do so. The sge scheduler doesn’t give an error when jobs time out.
* Move scheduler flags from the lesson body text to `statu_output.snip` because they can vary for different schedulers

### DIFF
--- a/_episodes/13-scheduler.md
+++ b/_episodes/13-scheduler.md
@@ -92,6 +92,7 @@ the *queue*. To check on our job's status, we check the queue using the command
 ```
 {{ site.host_prompt }} {{ site.sched_status }} {{ site.sched_flag_user }}
 ```
+{: .bash}
 {% include /snippets/13/statu_output.snip %}
 
 The best way to check our job's status is with `{{ site.sched_status }}`.
--- a/_episodes/13-scheduler.md
+++ b/_episodes/13-scheduler.md
@@ -66,6 +66,7 @@ run as a test.
 > echo 'This script is running on:'
 > hostname
 > sleep 120
+> echo 'This script has finished successfully'
 > ```
 > {: .bash}
 {: .challenge}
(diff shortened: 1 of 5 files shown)


_episodes/13-scheduler.md
```diff
@@ -91,15 +92,10 @@ the *queue*. To check on our job's status, we check the queue using the command
 ```
 {{ site.host_prompt }} {{ site.sched_status }} {{ site.sched_flag_user }}
 ```
-{: .bash}
```
tkphd, line 94: This `{:bash}` tag belongs to the preceding block: its deletion here is in error. Please restore.

_episodes/13-scheduler.md
```diff
@@ -135,12 +131,13 @@ Let's illustrate this by example. By default, a job's name is the name of the sc
 Submit the following job (`{{ site.sched_submit }} {{ site.sched_submit_options }} example-job.sh`):
 
 ```
-#!/bin/bash
+#!/bin/bash -l
```
tkphd, line 134: `-l` flag (run like a login shell) is extraneous, since none of the user's environment is relevant to the script.

_extras/snippets_library/EPCC_Cirrus_pbs/13/statu_output.snip
```diff
@@ -1,6 +1,11 @@
-
+{: .bash}
```
tkphd, line 1: This `{:bash}` tag belongs to the preceding block. Please remove from this snippet.

_includes/snippets/13/statu_output.snip
```diff
@@ -1,2 +1,8 @@
+{: .bash}
```
tkphd, line 1: This `{:bash}` tag belongs to the preceding block. Please remove from this snippet.

tkoskela:
Thank you for the reviews!

I moved the `{: .bash}` command back to the lesson markdown file, thanks for spotting that.

> Would it be possible to achieve the desired outcome (producing correct SGE output, with SGE-specific commentary) without moving the Markdown text into the code snippet?

The problem I was trying to address is, the run flags are different in SGE to slurm. For example, in slurm `R` is running while in SGE it is reserved, and the flag for running in SGE is `r`. Therefore, the two sentences that talk about the flags need to be in a snippet. Would you prefer I create a separate snippet for them, rather than putting them in `statu_output.snip`?

Maybe [our version of the snippet](https://github.com/UCL-RITS/hpc-intro/blob/gh-pages/_extras/snippets_library/UCL_sge/13/statu_output.snip) helps clarify this

tkphd:
Thanks for addressing these @tkoskela. Our line spacing makes it confusing which block the `{:.bash}` syntax flags attach to, so that's on us, too.

Looking through the lessons more closely, I believe that (1) we don't uniformly and (2) we should not be keeping Markdown separate from snippets. Doing so produces extraneous blank lines in a lot of places, and means the snippet files can't stand alone. I'm going to start working on overhauling the existing library to move the Markdown block definitions & syntax highlighting inside the snippet files. Please do as you see fit to achieve the desired formatting result.

psteinb:
Sorry, I think I triggered the conflicts by merging #127 
github doesn't allow me to make suggestions on how to resolve the conflicts. @tkoskela could you lend a hand?

tkphd:
I'm happy to fix the merge conflicts, since #128 is my doing. Note that this is easier to do using desktop tools (command line & three-way merge tools) than it is through the browser.

psteinb:
if @tkoskela approves, I think @tkphd is perfect fit for the task.

tkoskela:
Sounds good to me

psteinb:
@tkphd go for it. Thanks for your help in advance.

PR #134, opened 2020-07-07
Resolve vulnerability warning

tkphd:
Dependabot [warns](https://github.com/hpc-carpentry/hpc-intro/network/alert/Gemfile/jekyll/open) that Jekyll 3.4.0 is outdated. This change sets Jekyll at a minimum of 3.6.3, but allows a newer gem to be installed if available.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -2,7 +2,7 @@ source "https://rubygems.org"
 ruby RUBY_VERSION
 
 # This will help ensure the proper Jekyll version is running.
-gem "jekyll", "3.4.0"
+gem "jekyll", ">= 3.6.3"
 
 # Windows does not include zoneinfo files, so bundle the tzinfo-data gem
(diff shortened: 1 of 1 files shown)


psteinb:
I like the idea. To my surprise, e.g. [this lesson](https://github.com/swcarpentry/shell-novice) doesn't even have a `Gemfile`. Do you know why?
The carpentries workshop template does yield a `Gemfile` though: https://github.com/carpentries/workshop-template
I am confused.

tkphd:
We use Travis-CI to build and test our lessons -- as the Carpentries Template intends.
Many of the official Carpentries lessons have stripped this Continuous Integration functionality out, for unfathomable reasons.

It's possible that this vulnerability warning means we're out-of-step with the Carpentries Template, and a more careful synchronization is in order. I think that's a separate issue -- right now, I want to patch this vulnerability.

psteinb:
Sounds like a plan. Let's go for it.

PR #137, opened 2020-07-07
Sync up with Carpentries templates

tkphd:
This is a poorly-scoped PR, but broadly speaking, it brings HPC Intro up-to-date with the Carpentries [lesson-example](https://github.com/carpentries/lesson-example). This updates some syntax highlighting stuff, brings dependencies up to date, and makes some tweaks to the continuous integration scripts.

- Fixes line lengths, ignoring lines starting with "http" so long URLs are OK
- Adds HPC Carpentries logos, icons, and default URL for boilerplate HTML
- Overhauls the Setup page, making note of Powershell but preferring Git for Windows
- Makes note of how to include new figures (with Liquid variables)

Note that, relative to a pure sync with [carpentries/lesson-example](https://github.com/carpentries/lesson-example), this PR contains 1000 extra commits. This is because I merged [carpentries/workshop-template](https://github.com/carpentries/workshop-template) first, then found the appropriate repo and merged that, too. Undoing this saved very little, and resulted in a huge mess, and I did not want to try a third time around. So, please forgive my cluttering of the history and scope creep, but I think the current state is in better shape.


.mailmap
```diff
@@ -0,0 +1,20 @@
+Abigail Cabunoc Mayes <abigail.cabunoc@gmail.com>
```
psteinb, line 1: What is this file for?

psteinb:
merging this will surely produce conflicts with the other PRs. Should we risk that?

dpshelio:
Should we move to use carpentries-theme instead? it's a lot easier to manage and you don't get all these problems. 
https://github.com/carpentries/carpentries-theme
It only requires a line on the `_config.yml`.

.mailmap
dpshelio: `.mailmap` helps to get the contributors that uses different email as the same contributor. So if you filter git history by a name you get all the contributions that x person did with multiple names/emails.
https://git-scm.com/docs/git-check-mailmap

dpshelio:
If wanted I have a script (that I use for the translations) that will create a PR removing all the stuff that belongs to the theme...

tkphd:
@dpshelio please do!

tkphd:
Withdrawn. Let's migrate to [carpentries-theme](https://github.com/carpentries/carpentries-theme) instead, so we can focus on our corner of the world without worrying so much about upstream.

PR #139, opened 2020-07-08
Lean on the remote carpentries-theme

tkphd:
Jekyll supports remote themes, and The Carpentries [made one](https://github.com/carpentries/carpentries-theme). This PR replaces all the theme-related files in this repo with a one-line addition to `_config.yml`, and an updated `Gemfile`.

This is so much better than #137. Thanks for the tip, @dpshelio!


psteinb:
Happy to review or ask for reviews once this is ready. Thanks for this awesome contribution.

tkphd:
It is indeed a lot lighter!

I don't like that `make lesson-check` fails, and I'm worried that the links at the bottom of each page aren't working. I'll remove [WIP] and request new reviews soon.

tkphd:
OK -- following advice from @maxim-belkin on Slack, `make lesson-check-all` now works. If the footers fail, that's an issue to take up with the upstream carpentries-theme.

tkphd:
@dpshelio, is your script somewhere on GitHub already?

tkphd:
I merged this into my fork's `gh-pages` to get a rendered site. Please review it here: <http://www.trevorkeller.com/hpc-intro/>

I've found that the Carpentries links in the page footers are all broken, but the carpentries-theme link works fine. I'll submit an issue to request they look into this.

psteinb:
I clicked around in the rendered version and it looks super cool and stable to me. I saw the problems with the links to the carpentries as well. 
https://github.com/carpentries/carpentries-theme/issues/4
I'll merge this now in good hope that it will stream line this material and make our live easier.

bin/workshop_check.py
```diff
@@ -418,3 +418,4 @@ def main():
 
 if __name__ == '__main__':
     main()
+ain()
```
ocaisa, line 421: typo here I suspect

psteinb: https://github.com/hpc-carpentry/hpc-intro/pull/144

dpshelio:
> @dpshelio, is your script somewhere on GitHub already?

@tkphd Sorry I missed this! Yes it is, but it's to a slightly different theme, one with multilanguage support:
https://github.com/carpentries-i18n/i18n/blob/master/helpers/lesson2theme.py
It does too many things (I should separate the bits a bit more), but the function you would want is themetise_lesson`](https://github.com/carpentries-i18n/i18n/blob/master/helpers/lesson2theme.py#L78)

tkphd:
Cool -- this would set `remote: carpentries-i18n/carpentries-theme`, right?
That's probably a switch we should make here as well, given the international (i11l?) nature of the HPC Carpentry crew.

... how would I properly invoke `themetise_lesson`? Like this?
```bash
$ cp ${SOMETHING}/helpers/lesson2theme.py ./
$ python
>>> from lesson2theme import clone_repo_as_submodule, themetise_lesson
>>> org = "hpc-carpentry"
>>> repo = "hpc-intro"
>>> parent, clone, upstream, fork, submod, github_user = clone_repo_as_submodule(org, repo, tmpdir="/tmp")
>>> themetise_lesson(clone)
>>> quit()
$ cd /tmp/hpc-intro
$ bundle exec jekyll serve
```

psteinb:
Thanks to both of you @dpshelio and @tkphd, preparing for i18n (once I understood the meaning of the `18`, I started to love writing this abbreviation) is key in my view to spread and disseminate this material. Please keep going.

dpshelio:
so... for the i18n this needs a bit of work. The theme is not 100% working yet (as you could see on [some issues I've encountered](https://github.com/carpentries-i18n/carpentry-theme/issues?q=is%3Aissue+is%3Aopen+sort%3Aupdated-desc)).

At the moment, that script, yes only adds the carpentries-theme for i18n, but it's not that one... yet, it's `carp-theme` because I had to tests many things. I'll transition to the good one sometime soon (hopefully before my talk at the end of August). Your snippet seems right... but right now, you just need the change the theme.

To benefit of the whole translations things... for this type of lesson... I still don't know how to do it because of the snipets. But I'm happy to discuss possibilities :)

PR #141, opened 2020-07-10
Suspected typo

ocaisa:
from #139

### DIFF
--- a/bin/workshop_check.py
+++ b/bin/workshop_check.py
@@ -418,4 +418,3 @@ def main():
 
 if __name__ == '__main__':
     main()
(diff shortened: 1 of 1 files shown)


psteinb:
Question is, what is right omitting the main call as in this PR or having it as in #144 ?

ocaisa:
> Question is, what is right omitting the main call as in this PR or having it as in #144 ?

I was going off what is in https://github.com/carpentries/styles/blob/gh-pages/bin/workshop_check.py

psteinb:
Thanks for the fix.

PR #142, opened 2020-07-10
fix footer links with more recent list of links

fmichonneau:
this should fix carpentries/carpentries-theme#4

The footer relies on variables defined in the `_config.yml` file that were not present leading to the broken links.


tkphd:
Yup, that'll do it. Thanks!

PR #143, opened 2020-07-10
Ignore .vendor dir which houses bundle and...

ocaisa:
don't use deprecated bundle path option


tkphd:
Please also make a PR against the upstream Carpentries [lesson-example](https://github.com/carpentries/lesson-example). Thanks!

PR #145, opened 2020-07-10
Improve description of terms

markcmiller86:
I took a stab at improving the description of terms here. It results in a slightly longer introductory paragraph but I think the clarity it provides (at least I hope) could be worth the extra space.

### DIFF
--- a/_episodes/12-cluster.md
+++ b/_episodes/12-cluster.md
@@ -29,7 +29,9 @@ Brazil, Zimbabwe and Japan may manage those resources as its own internal cloud 
 company may also utilize commercial cloud resources provided by Amazon or Google. Cloud resources
 may refer to machines performing relatively simple compute tasks such as serving websites, providing
 shared storage, providing webservices (such as e-mail or social media platforms), as well as more
-traditional compute intensive tasks such as running a simulation. An *HPC system* on the other hand,
+traditional compute intensive tasks such as running a simulation.
+
+An *HPC system* on the other hand,
 is a term used to describe individual computing systems, typically with many, monolithic processing
 and storage elements, designed and exclusively used for compute intensive work. For example, all
 of the computing systems on the [Top-500](https://www.top500.org) list are HPC systems. HPC systems
--- a/_episodes/12-cluster.md
+++ b/_episodes/12-cluster.md
@@ -39,7 +39,8 @@ of the machines on the [Top-500](https://www.top500.org) list are HPC systems. T
 an HPC resource must exist in a specific, fixed location: networking cables can only stretch so far, and
 electrical and optical signals can travel only so fast.
 
-The word "cluster" is often used for small to moderate scale HPC resources less impressive than the Top
+The word "cluster" is often used for small to moderate scale HPC resources less impressive than the [Top-
+500](https://www.top500.org). Clusters are often maintained in computing centers that support several such systems, all sharing
 500. Clusters are often maintained in computing centers that support several such systems, all sharing
 common networking and storage to support common compute intensive tasks.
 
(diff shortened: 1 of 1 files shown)


tkphd:
Thanks for the contribution, @markcmiller86. There is healthy & ongoing debate over the "definitions" of cloud and HPC, and this text is a welcome addition.

I think that the goal of this introductory paragraph was to provide a high-level overview of the kinds of things these resources (cloud vs. HPC) can do. In order to maintain its scope and feel, I recommend moving your explanations into the glossary in [reference.md](https://github.com/hpc-carpentry/hpc-intro/blob/gh-pages/reference.md) and linking there for interested readers.

Contrasting opinions are always welcome :-)

markcmiller86:
I know I am late to the party. And, in all likelihood, given my schedudle, this might be the only PR I submit for this lesson.

If you can suggest some revision of the text I provided that reduces length but maintains meanings, or otherwise addresses any *specific* concerns you have with it, that would be great.

I don't want to have a debate either. But, I don't  think this lesson or its authors should place themselve in the position of declaring that the community uses the terms *cloud*, *cluster* or *HPC* with *"...varying degress of correctness*" as that alone suggests there is indeed a *correct* outcome to such a debate. And, FWIW, `cloud != remote`. Remote is part of cloud but actually probably the least important and so not a way to distinguish it or something to emphasize about it.

All that said, the suggestion to move these *introductory* remarks from where the terms are *first* being used to a *back-page* that we all know no one will ever read doesn't feel like a good compromise either.

tkphd:
I didn't mean "debate" as a negative thing, more of a discussion to convey clearly & concisely as possible to learners what we mean. There is no "right way" to explain these concepts, and I suspect we'll continue to iterate. Your point, that this is the "What do we mean by HPC?" section, is very well taken. I'll take another look, but this looks pretty good to me.

I hope this is not your last PR here. Your contribution is appreciated, and more perspectives always helps improve the outcome.

tkphd:
I like your overview of the cloud. The subtle distinction that a Cloud is not necessarily somebody else's computer is a good one. The narrow qualifiers of provisioning and potential virtualization are good.

The HPC section is less clear. It starts by defining HPC as an *individual system*, but closes by differentiating a *cluster* as a "computing center" where several HPC systems may be tightly integrated. My understanding is that a cluster can be any networked collection of computers, from a pair of Raspberry Pis all the way up to Fugaku. Only one of those is a Top 500 machine.

Please consider the following:

    The term *HPC system*, on the other hand, describes a stand-alone resource for computationally intensive workloads.
    They are typically built from numerous independent processing and storage elements, designed to handle high
    volumes of data, large numbers of mathematical operations, or both, with the lowest possible latency. For example, all
    of the machines on the [Top-500](https://www.top500.org) list are HPC systems. In order to support these constraints,
    an HPC resource must exist in a specific, fixed location: networking cables can only stretch so far, and electrical and
    optical signals can travel only so fast. The word "cluster" is often used for HPC resources less grandiose than the 
    Top 500. Clusters are often maintained in computing centers that support several such systems, all sharing common 
    networking and storage to support common compute intensive tasks.

_episodes/12-cluster.md
```diff
@@ -18,16 +18,31 @@ keypoints:
...
+
+The term *HPC system*, on the other hand, describes a stand-alone resource for computationally
+intensive workloads. They are typically comprised of a multitude of independent processing and storage
+elements, designed to handle high volumes of data and/or large numbers of floating-point operations
+([FLOPS](https://en.wikipedia.org/wiki/FLOPS)) with the highest possible performance. For example, all
+of the machines on the [Top-500](https://www.top500.org) list are HPC systems. To support these constraints,
+an HPC resource must exist in a specific, fixed location: networking cables can only stretch so far, and
+electrical and optical signals can travel only so fast.
+
+The word "cluster" is often used for small to moderate scale HPC resources less impressive than the Top
```
psteinb, line 42:
```suggestion
The word "cluster" is often used for small to moderate scale HPC resources less impressive than the [Top-
500](https://www.top500.org). Clusters are often maintained in computing centers that support several such systems, all sharing
```

PR #149, opened 2020-07-13
Clear lesson check issues

tkphd:
`make lesson-check-all` warns about numerous files containing excessively long lines.
Some are URLs or templated commands, which are best left alone. Most are, indeed,
quite long. This PR excuses the former, and addresses the latter.

### DIFF
--- a/README.md
+++ b/README.md
@@ -13,8 +13,8 @@ writer, and then reviewed by the rest of the group once complete.
 
 ## Using this material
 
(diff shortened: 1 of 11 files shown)


ocaisa:
We should probably document this somewhere for how you are handling the links as the syntax is pretty particular about where the URL needs to be.

Another option would be to exclude lines with a url in general:
```
and (not 'http' in l)
```

tkphd:
Great idea, @ocaisa! Searching for substrings, rather than matching specific line beginnings, is a much more forgiving & less tedious approach. Thanks for the suggestion!

PR #153, opened 2020-07-17
Name to job ~and change order of commands~

dpshelio:
It would make a good practice to name the job so it's easier to find it on the queue.
Also, it makes more sense to me that the `echo` and `hostname` are printed together before `sleep` starts.

### DIFF
--- a/_episodes/13-scheduler.md
+++ b/_episodes/13-scheduler.md
@@ -251,11 +251,12 @@ minutes.
 
 ```
 #!/bin/bash
+{{ site.sched.comment }} {{ site.sched.flag.name }} long_job
 {{ site.sched.comment }} {{ site.sched.flag.time }} 00:00:30
 
 echo -n "This script is running on "
-sleep 120 # time in seconds
 hostname
+sleep 120 # time in seconds
 echo "This script has finished successfully."
 ```
 {: .output}
--- a/_episodes/13-scheduler.md
+++ b/_episodes/13-scheduler.md
@@ -254,7 +254,7 @@ minutes.
 {{ site.sched.comment }} {{ site.sched.flag.name }} long_job
 {{ site.sched.comment }} {{ site.sched.flag.time }} 00:00:30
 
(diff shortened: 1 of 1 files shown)


tkphd:
Having just introduced job names, we should definitely include that in this script and every subsequent one. 

Wearing my programmer hat, I agree that it makes little sense to set a thought aside and sleep for a bit.
With my teacher hat on... We're trying to construct a script that runs afoul of the time limit, and makes it obvious it has done so. Adding a clear statement like "This script has completed." works, but for someone *only* examining the output, missing that last sentence would not give the impression anything has gone awry. Having only the start of the sentence, however, makes it immediately obvious that something did not go to completion.

Probably the best answer is to make the script longer, with clear pre-sleep and post-sleep segments, along the lines of

```bash
#!/bin/bash
#SBATCH -t 00:00:30
#SBATCH -J drowsy-node

echo "Hello! I'm... drowsy..."
sleep 120 # time in seconds
echo -n "Oh! Yes, I'm "
hostname
```

dpshelio:
Got-ya!
I thought it has been a typo. I added a `...` to the `echo` line to make it clearer that we are "waiting".

_episodes/13-scheduler.md
```diff
@@ -251,9 +251,10 @@ minutes.
 
 ```
 #!/bin/bash
+{{ site.sched.comment }} {{ site.sched.flag.name }} long_job
 {{ site.sched.comment }} {{ site.sched.flag.time }} 00:00:30
 
-echo -n "This script is running on "
+echo -n "This script is running on ..."
```
tkphd, line 257: Using `echo -n`, the hostname will print immediately after the `...`
Please add a space after the final `.` so we get `This script is running on ... landing_pad` instead of `This script is running on ...landing_pad`. Thanks!

tkphd:
Sorry to revert approval -- just a minor change for a missing space. Thanks for your flexibility @dpshelio!

PR #163, opened 2020-07-17
rename snippets to be less confusing

### LINKED ISSUE #151: snippet names are confusing
It's very difficult to track down which snippet one must edit to port or correct the output. Partially, this stems from reuse of "example_job.sh", requiring the lesson to use the contents of that file as it changes to specify snippets.

A better approach would be to write one shell script per job, copying old ones where appropriate, and using the script name to inform the snippet name. If we're very clever, the naming could escalate alphabetically...

tkphd:
The original names of the snippets don't provide enough information to guess at their contents with any success.
This PR closes #151 by renaming the snippets in more verbose terms, with names chosen such that the lessons proceed through them in alphabetical order. This helps to clarify, in at least one person's brain, the structure and flow of the libraries.

In addition, a utility (`rename-snippets.sh`) has been included. When supplied with a folder name containing snippets in the original layout, it uses `git mv` to rename them to match this format. Extra snippets, created for individual sites, remain unaffected.


psteinb:
Thanks for this PR. I wonder though, should we introduce a differentiation between input snippets and output snippets?

tkphd:
I had something like that in mind, but beyond the most simplistic snippets, there's no clean delineation: several large snippets include input, output, and lesson material. I think the cleanest approach is to keep the basic structure as-is, and iterate on naming schemes until we find something we can be proud of.

PR #104, opened 2019-09-26
Suggest more customization using site variables [WIP]

Sabryr:
Suggest more customization after using the material with other cluster with differing setups. 

1. include site.ls_user_home_output, so different ls out puts can be showing for different cluster setups Fixes #94

PR #169, opened 2020-07-27
What's in your home directory?

### LINKED ISSUE #94: Make Working on a cluster more universal

tkphd:
Introduce an exercise using `pwd`, `ls`, and `df` to get familiar with the remote environment.

Closes #94. Supersedes #104.


reid-a:
In the first paragraph, "folder" should be plural, to be grammatically parallel with "files" and "links".

On line 116 of the diff, "yourUsername" is literal. I think the general direction is to do these from the template data?

tkphd:
Fixt, thanks for the pointers!

Issue #173, opened 2020-07-29
How are multiple  _config.yml maintained 


Sabryr:
At the moment I need to stash my _config.yml during push/pulls. Wondering how others are handling it (may be there is an elegant way) . Then what is the use of (where is it useed),
_includes/snippets_library/ComputeCanada_Graham_slurm/_config_options.yml

IMO, most of the _config.yml could be transferred to snippets as well

reid-a:
I believe the thinking behind the current design is that the 
_config.yml is a convenient central location for variables that
any given site-specific lesson instance will certainly need.

  That it would require customization is OK, in this model, 
because it's not conceptually different from things like the 
venue location and date and other instance-specific data you'd 
have to customize anyways before publishing an announcement.

  But on reflection, it's true that site data is longer-lived
than instance data.  On the other hand, the items you'll need
(but not their values) are common across all sites, so what
we really want is to centralize the structure of the configuraiton,
but distribute the site-specific values.

  Not sure how to do that.

 				-- A. 

On Wed, Jul 29, 2020 at 01:44:43AM -0700, Sabry Razick wrote:
> At the moment I need to stash my _config.yml during push/pulls. Wondering how
> others are handling it (may be there is an elegant way) . Then what is the use
> of (where is it useed),
> _includes/snippets_library/ComputeCanada_Graham_slurm/_config_options.yml
> 
> IMO, most of the _config.yml could be transferred to snippets as well
> 
> —
> You are receiving this because you are subscribed to this thread.
> Reply to this email directly, view it on GitHub, or unsubscribe.*
> 

-- 
Dr. Andrew C. E. Reid
Physical Scientist, Computer Operations Administrator
Center for Theoretical and Computational Materials Science
National Institute of Standards and Technology, Mail Stop 8555
Gaithersburg MD 20899 USA
andrew.reid@nist.gov

tkphd:
Excellent point, @Sabryr, and one that's important for us all to be aware of. You're right that, at this point, changes to `_config.yml` will get clobbered when you `git pull` from hpc-carpentry. The purpose of the `_config_options.yml`s in the snippet library is to maintain your site-specific settings, so that they can be restored to `_config.yml` when it's overwritten, or for someone attempting to duplicate (or branch off from) your site to get those settings right. 

[...]

PR #177, opened 2020-07-30
Use environment variables to set site config

tkphd:
Jekyll allows for multiple config files (e.g., `_config.yml`) on the command line, using the `--config` flag. When multiple files are specified, any variables defined in both assume the values in the last file specifying them.

This PR sets the Travis-CI script to test each of the known clusters. The `Makefile` is altered to set `--config _config.yml,` by default; extra files can be appended through the Make variable `SITE_CONFIG`. Each of these configurations (default and with the specified configs) is then test-built, to check for Jekyll-breaking errors in the snippet files.

This represents an improvement in portability: rather than *overwriting* `_config.yml`, we can simply specify an additional config file, and no changes to Git-tracked files are necessary.

Addresses #173.


psteinb:
@tkphd I saw commits coming in to this PR. Is this still WIP or do you consider it ready for review?

tkphd:
It's ready for review.

PR #190, opened 2020-08-04
Speed up travis

### LINKED ISSUE #189: Travis builds are entirely too slow
At present, Travis spins up a separate build for each of the known sites, creating a neat dashboard of build successes (and failures). Unfortunately, this is done in serial, and the package installation from one does not carry over to the next. As a consequence, each config takes 5 minutes to test, and the total build takes up to 30 minutes to complete. This is tragic.

Recommended fix: rewrite the script to have just two builds, Spell Check and Test, and run the "make site" script within just that one, not multiple, builds.

tkphd:
Fixes #189

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,6 @@
 # Travis-CI config for https://github.com/hpc-carpentry/hpc-intro
 # Results at https://travis-ci.org/github/hpc-carpentry/hpc-intro
-
+---
 dist: xenial
 language: ruby
 python: 3.7
@@ -8,8 +8,8 @@ cache: bundler
 
 branches:
   only:
-  - gh-pages
-  - /.*/
+    - gh-pages
+    - /.*/
 
 before_install:
   - sudo apt-get update -y
@@ -19,7 +19,7 @@ before_install:
   - bundle install
 
 install:
- pip install codespell pyyaml
+  pip install codespell pyyaml
 
 script:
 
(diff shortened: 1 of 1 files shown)


reid-a:
Is there a danger that this will lead to "stealth breakage" that goes undetected until someone actually tries to deploy their local lesson? 

The candidate scenario is, someone makes a lot of changes to their local configuration set-up, and it builds locally on their laptop, but drifts away from what the CI is expecting, and the CI doesn't catch it. Then, when that site goes to actually stand up a lesson "for real", the CI issues have to be debugged on a short time-scale at the same time as the rest of the lesson logistics.

If the configs are largely settled, or if people's laptops and the CI build environment are close enough that this type of divergence is unlikely, then the risk is low.

tkphd:
Not sure I follow. Every site in the snippets library still gets tested, and if one fails, the build will halt and report that failure. The only difference is that, rather than calling out each build separately, they're bundled together into one job on Travis. Well, two -- it builds the "default" site as one job, and all of the specific libraries as another. 

So... in terms of stealth breakage, this PR shouldn't change that. And if you're running some version of `make site` locally, deploying with gh-pages should work fine, provided you set the `SITE_CONFIG` environment variable or provide it as demonstrated in `.travis.yml`.

reid-a:
OK, thanks. I misread the .diff, I thought it was down-scoping, but it's just de-serializing.

PR #193, opened 2020-08-05
Add GitHub Actions workflow

ocaisa:
This adds a GitHub Actions workflow for CI, which we can ultimately use to replace Travis. The jobs are run in parallel so CI time is about 5-6 minutes. Actually building the lesson is trivial (seconds) so there isn't really much point to the parallelism, we could just reuse the environment from one case. The only thing is how easy it would be to add additional sites with the way I've implemented it.

### DIFF
--- a/.github/workflows/test_and_build.yml
+++ b/.github/workflows/test_and_build.yml
@@ -47,7 +47,12 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        SITE_CONFIG: [ComputeCanada_Graham_slurm, EPCC_Cirrus_pbs, Norway_SIGMA2_SAGA_slurm, UCL_Myriad_sge]
+        SITE_CONFIG:
+          - ComputeCanada_Graham_slurm
+          - EPCC_Cirrus_pbs
+          - NIST_CTCMS_slurm
+          - Norway_SIGMA2_SAGA_slurm
+          - UCL_Myriad_sge
     steps:
     - uses: actions/checkout@v2
     - uses: actions/setup-ruby@v1
--- a/.github/workflows/test_and_build.yml
+++ b/.github/workflows/test_and_build.yml
@@ -5,6 +5,7 @@ on:
   push:
     branches:
       - gh-pages
+      - github_actions
 
 jobs:
   spellcheck:
--- a/.github/workflows/test_and_build.yml
+++ b/.github/workflows/test_and_build.yml
@@ -1,0 +1,65 @@
+name: Check lesson and build for all configs
+
+on:
+  pull_request:
+  push:
+    branches:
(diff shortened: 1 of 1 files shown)


ocaisa:
See the result from the commit here at https://github.com/ocaisa/hpc-intro/actions/runs/196285918

ocaisa:
It's not really the image that's the problem, the gem stuff related to ruby (which I don't really understand) takes all the time.

I tried a switch to `alpine-latest` and [that didn't work](https://github.com/ocaisa/hpc-intro/actions/runs/196309974), the actual options here are very limited https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#jobsjob_idruns-on

.github/workflows/test_and_build.yml
```diff
@@ -0,0 +1,65 @@
...
+        make lesson-check-all
+    - name: "Check build"
+      run: |
+        make --always-make site
+
+  build-specific-sites:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        SITE_CONFIG: [ComputeCanada_Graham_slurm, EPCC_Cirrus_pbs, Norway_SIGMA2_SAGA_slurm, UCL_Myriad_sge]
```
tkphd, line 50: * Array missed an entry, `NIST_CTCMS_slurm`
* Having these all on one line is hard to parse & maintain. Please try the alternative syntax for a YAML array,
  ```yaml
  matrix:
    SITE_CONFIG:
      - ComputeCanada_Graham_slurm
      - EPCC_Cirrus_pbs
      - NIST_CTCMS_slurm
      - Norway_SIGMA2_SAGA_slurm
      - UCL_Myriad_sge
  ```

psteinb: Like so, I guess:
```suggestion
        SITE_CONFIG:
    - ComputeCanada_Graham_slurm
    - EPCC_Cirrus_pbs
    - NIST_CTCMS_slurm
    - Norway_SIGMA2_SAGA_slurm
    - UCL_Myriad_sge
```

tkphd: Looks like you need 6 more spaces at the beginning of each line.

tkphd:
@ocaisa does the modified array work on Actions, or will that not be tested before merging?

ocaisa:
Testing in https://github.com/ocaisa/hpc-intro/actions/runs/196513209 which parsed correctly, will revert that commit once those tests pass

tkphd:
GitHub Actions [failed](https://github.com/ocaisa/hpc-intro/actions/runs/196513209) because the CTCMS snippet library does not exist on @ocaisa's fork, but it does here, and everything else passed. Merging.

PR #194, opened 2020-08-05
replace Travis with GitHub Actions, incl. badge & README notes

tkphd:
Building with Travis takes too long, and is redundant with GitHub Actions. This PR completes the transition started in #193.


tkphd:
These build checks are delightfully quick. Thanks for porting us over, @ocaisa!

PR #203, opened 2020-08-07
Update index.md

annajiat:
1. added alternative terms:  shell (also known as terminal or command line)
2. fixed layout/callout of pre-requisites, was mixed with getting started. 
3. added URL/link text for shell-novice
hpc-intro-changes

### DIFF
--- a/index.md
+++ b/index.md
@@ -19,8 +19,8 @@ By the end of this workshop, students will know how to:
 >
 > To get started, follow the directions in the "[Setup](setup/)" tab to 
 > download data to your computer and follow any installation instructions.
->
->
+{: .callout}
+
 > ## Prerequisites
 >
 > Command line experience is necessary for this lesson. We recommend the participants to go through
--- a/index.md
+++ b/index.md
@@ -10,7 +10,7 @@ overview of the tools available and how to use them effectively.
 
 By the end of this workshop, students will know how to:
 
-* Use the UNIX command line to operate a computer, connect to a cluster, and write simple shell
+* Use the UNIX shell (also known as terminal or command line) to operate a computer, connect to a cluster, and write simple shell
   scripts.
 * Submit and manage jobs on a cluster using a scheduler, transfer files, and use software through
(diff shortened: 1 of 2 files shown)


tkphd:
Thank you for the contribution @annajiat!
The text changes look good, helpfully including synonyms for the shell in case it's unclear. Pulling the Prereqs section from header level 4 to 2, inline with the rest, is also useful. That looks good.

You've also stripped the quotation block highlighting flags, which is less good. The site now renders as below, appearing much less polished than before. Please restore those helpful cues.

index.md
```diff
@@ -20,19 +20,22 @@ By the end of this workshop, students will know how to:
...
 >
-> #### Prerequisites
+>
+> ## Prerequisites
 >
 > Command line experience is necessary for this lesson. We recommend the participants to go through
-> <https://swcarpentry.github.io/shell-novice/> if new to the terminal.
-{: .prereq}
+> [shell-novice](https://swcarpentry.github.io/shell-novice/), if new to the command line (also known as terminal or shell).
+> {: .prereq}
```
tkphd, line 28: Blockquote types are specified after and outside, not within, the blockquote. Please remove the extraneous `> ` from each of these.

annajiat: @tkphd 
Thank you for the the guidance.

1.  Could you please check if this one is correct?
https://github.com/hpc-carpentry/hpc-intro/commit/4b9c983046f7e581398e75e45444ed0ceebd2c34

2. I have removed the fork as per suggestions by git that it is no longer needed. Was it incorrect? Should I start a new PR?

3. Those were originally (and incorrectly) removed as they showed up as a part of the text in the git preview. Should that be ignored? Am I missing anything else here?

tkphd: The preview on GitHub does not include the Carpentries theme. To preview the pages as they will render on the Web, you will have to build the modified site locally. Some directions can be found [here](https://carpentries.github.io/lesson-example/setup.html).

PR #206, opened 2020-08-07
See if we can shorten CI time

ocaisa:
I noticed that we are installing 2 different versions of Jekyll and I was wondering if we can drop one of these.

### DIFF
--- a/.github/workflows/test_and_build.yml
+++ b/.github/workflows/test_and_build.yml
@@ -33,7 +33,7 @@ jobs:
       run: |
         # Need this library for nokogiri
         sudo apt-get install libxslt1-dev    
-        gem install bundler
+        gem install bundler json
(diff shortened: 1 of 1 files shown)


psteinb:
the code changes look ok, but the tests appear to fail.

.github/workflows/test_and_build.yml
```diff
@@ -33,9 +33,10 @@ jobs:
       run: |
         # Need this library for nokogiri
         sudo apt-get install libxslt1-dev    
-        gem install bundler jekyll json kramdown
+        gem install bundler
```
tkphd, line 36: restore json and kramdown here?

tkphd: 8de7fc4

PR #207, opened 2020-08-07
Remove copy/paste bug in lesson initialisation

ocaisa:
See https://github.com/carpentries/styles/blob/gh-pages/bin/lesson_initialize.py


tkphd:
Build error is unrelated.

PR #212, opened 2020-08-07
use HPC_JEKYLL_CONFIG to switch sites

### LINKED ISSUE #195: SITE_CONFIG is a vague env var name
A more specific name would be preferable, when telling Make which site to build.

tkphd:
Closes #195.


tkphd:
Preceded by #211!

PR #217, opened 2020-08-13
Address review comments on lesson 13

### LINKED ISSUE #181: Review comments on lesson 13
- [ ] Restaurant analogy is awkward, leaves out "chefs" and "waiters" 
      which are the more obvoius/intuitive workers. 
- [x] Shell-script example introduces creation of the script,
      but depicts the output of listing the script, not the creation process.
- [X] First sbatch example fails on systems with no default partiiton.
- [x] E-mail notification call-out should have a solution.
- [x] Resource req's -- "-N" is nodes, discussed earler. -n is ranks/tasks.
- [x] Job environment variables does not have a solution.
- [x] Again with the cpu, in the interactive part.

tkphd:
Resolves most of the issues raised in #181, though the restaurant queue item is left untouched.

Closes #181.

### DIFF
--- a/_includes/snippets_library/ComputeCanada_Graham_slurm/13/print-sched-variables.snip
+++ b/_includes/snippets_library/ComputeCanada_Graham_slurm/13/print-sched-variables.snip
@@ -1,13 +1,29 @@
 > ## Job environment variables
 >
-> When SLURM runs a job, it sets a number of environment variables for the job. One of these will
-> let us check our work from the last problem. The `SLURM_CPUS_PER_TASK` variable is set to the
-> number of CPUs we requested with `-c`.
->
-> Using the `SLURM_CPUS_PER_TASK` variable, modify your job so that it prints how many CPUs have
-> been allocated.
+> When {{ site.sched.name }} runs a job, it sets a number of environment variables for the job.
+> One of these will let us check what directory our job script was submitted from.
+> The `SLURM_SUBMIT_DIR` variable is set to the directory from which our job was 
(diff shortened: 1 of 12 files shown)


_includes/snippets_library/ComputeCanada_Graham_slurm/13/print-sched-variables.snip
```diff
@@ -1,13 +1,29 @@
...
 >
-> When SLURM runs a job, it sets a number of environment variables for the job. One of these will
-> let us check our work from the last problem. The `SLURM_CPUS_PER_TASK` variable is set to the
-> number of CPUs we requested with `-c`.
->
-> Using the `SLURM_CPUS_PER_TASK` variable, modify your job so that it prints how many CPUs have
-> been allocated.
+> When {{ site.sched.name }} runs a job, it sets a number of environment variables for the job.
+> One of these will let us check what directory our job script was submitted from.
+> The `SLURM_SUBMIT_DIR` variable is set to the directory from which our job was
```
reid-a, line 5: The scheduler has been abstracted out to a liquid variable, but the actual environment variables in the text are still literals, and are SLURM-specific.

tkphd: Since this is a snippet in the ComputeCanada SLURM cluster library, I'm comfortable leaving these hard-coded. I'm skeptical that the variable for `SLURM_SUBMIT_DIR` would be used anywhere outside of this challenge problem, and don't want to litter the config files.

PR #225, opened 2020-08-19
Issue 214 parallel example

### LINKED ISSUE #214: Parallel example to run?
The death of seqtk reminds me that there's really nowhere in the lesson where a parallel task is executed by the learners.

I'm sympathetic to the argument (which I think I myself have made in the past) that this is impractical because MPI codes in general are not portable, so to have an example you can run you'd have to build it, and introducing building things is a large can of worms, and varies widely across systems.

Maybe somebody else has an idea of something learners could run that would demonstrate parallel processing?

For testing on my own clusters, I often use a simple "cpi.c" series expansion code that computes pi by a distributed series summation, borrowed from "examples" directory of the OpenMPI Debian package.

Feedback on whether this is too confusing or tangential would be welcome.

I was going to attach the file, but the issue template doesn't recognize the `.c` file type, so it'll be in-line in the first comment.

reid-a:
Closes #214. May have parameter breakage in non-slurm clusters.


tkphd:
I propose moving this from Ep. 14 (Modules) to Ep. 15 (Transferring). Since it uses both `python` and `wget`, both lessons are prerequisites, and the example code applies to any cluster where Python is installed, not just those using modules for software management.

reid-a:
I like the idea of a separate lesson. The scope was chosen because it's where the "setqk" example used to be, and this is kind of a more sophisticated version of that, or at least that's how the conversation started.

I think a lesson between the current 15 and 16 would be better.

tkphd:
make-it-so

tkphd:
Downside: the Snippet Libraries have directories hard-coded with lesson numbers, which are now off-by-one for $n > 15$. Recommend a separate issue to resolve that, perhaps in a more-maintainable way.

reid-a:
Ah, indeed. It's not broken, it's just that e.g. the now-lesson 17 has a reference to snippets/16/account-history.

More urgent, IMO, is that the PBS-alikes (Torque and SGE) should have a look at what was done for their node and task counts, and weigh in on how far off-base it is.

tkphd:
Agreed. Do we have access to one, so we can test internally?

reid-a:
I think there's a problem here.  Torque (and SGE?) specify nodes and "tasks" with `-l nodes=<N>:ppn=<n>`, whereas SLURM does it by `-N <N> -n <n>`.

Here is my proposed solution:

In the config, have site.sched.flag.nodetask1 and site.sched.flag.nodetask2.  Restrict it to a fixed number of tasks, for simplicity. For Torque and SGE, nodetask1=`-l nodes=1:ppn=4` and nodetask2 is blank. For SLURM, nodeask1=`-N 1` and nodetask2 is '-n 4`.

Then use nodetask1 and nodetask2 on successive lines in the output snippet.

You get a script that will run in either case, at the risk of a spurious blank line in the Torque one.

tkphd:
What if -- hear me out, it makes me squirm too -- we don't set *every* flag?

Slurm doesn't need `-N` unless you (a) want to overcommit, `-N=1 -n 128 --overcommit` or (b) specifically want to under-saturate nodes, `-N 2 -n 4`. If PBS-alikes are also reasonable bright, we can save a lot of explanation and config by not setting `-nnodes`. We only ever use one node anyway, so... ?

reid-a:
Torque experts can correct me if appropriate, but it's not obvious that one can set `ppn` in isolation in Torque. `-l ppn=4` is an error in my test script.

reid-a:
Alternate solution: Make the task specification itself a snippet, and incorporate it into the described run-file. This is only one step away from making the run-file itself a snippet, which I have been resisting, because it duplicates some info, especially the comment prefix, `#SBATCH` (or `#PBS` or the gridengine one), but it feels like it's in nearly the right scope, and solves the one-line vs two-lines problem.

tkphd:
"pll"?

reid-a:
Negotiable. For "parallel", the idea being this is the thin edge of the wedge of changing the snippet subdirectories to having names.

In my defense, it's infinitely more descriptive than "16".

tkphd:
Ah.

reid-a:
Also, I have just now verified on my system that it builds the PBS and SGE files in the expected way.

reid-a:
The "merge" operation to e1c5022 referred to above has no content changes, it's the result of rebasing to accommodate changes from PR #222 (and possibly #215, #217, #218, #220 -- the "review" PRs), since Github thought there were conflicts, although no conflicts arose in the rebasing process.

Issue #226, opened 2020-08-20
Rename snippet library sub-directories


tkphd:
At present, each of the sub-directories of each site's snippet catalog (to extend the library metaphor) is named with the numerical value of the lesson it goes with. However, as we see in #225, when new material is inserted, these numbers go out of sync. This is nightmarish to maintain, especially given the number of forks and pull requests that would need updating any time the lesson order changed.

Recommend finding a content-based naming scheme, with some verbosity, to improve maintainability & legibility. Related to #9.

## Current

| File Name                | Title                              | Snippet Folder Name |
| ---                      | ---                                | ---                 |
| 11-hpc-intro.md          | Why use a Cluster?                 | 11                  |
| 12-cluster.md            | Working on a remote HPC system     | 12                  |
| 13-scheduler.md          | Scheduling jobs                    | 13                  |
| 14-modules.md            | Accessing software                 | 14                  |
| 15-transferring-files.md | Transferring files                 | 15                  |
| 16-resources.md          | Using resources effectively        | 16                  |
| 17-responsiblity.md      | Using shared resources responsibly | 17                  |

## PR #225

| File Name                | Title                              | Snippet Folder Name |
| ---                      | ---                                | ---                 |
| 11-hpc-intro.md          | Why use a Cluster?                 | 11                  |
| 12-cluster.md            | Working on a remote HPC system     | 12                  |
| 13-scheduler.md          | Scheduling jobs                    | 13                  |
| 14-modules.md            | Accessing software                 | 14                  |
| 15-transferring-files.md | Transferring files                 | 15                  |
| 16-parallel.md           | Running a parallel job             | pll                 |
| 17-resources.md          | Using resources effectively        | 16                  |
| 18-responsiblity.md      | Using shared resources responsibly | 17                  |

## Proposed

| File Name                | Title                              | Snippet Folder Name |
| ---                      | ---                                | ---                 |
| 10-hpc-intro.md          | Why use a Cluster?                 | intro               |
| 15-cluster.md            | Working on a remote HPC system     | cluster             |
| 20-scheduler.md          | Scheduling jobs                    | scheduler           |
| 25-modules.md            | Accessing software                 | modules             |
| 30-transfers.md          | Transferring files                 | transfers           |
| 35-parallel.md           | Running a parallel job             | parallel            |
| 40-resources.md          | Using resources effectively        | resources           |
| 45-responsiblity.md      | Using shared resources responsibly | responsibility      |

[...]

PR #229, opened 2020-09-23
Renaming directories.

reid-a:
This commit does the directory-renaming part of issue #226, but
retains the sequential integer prefix for the episode file names.

Modifying or removing the numerical prefix involves changes to
the style code (so tests will still work), and possibly interacting
with the upstream Carpentries repo, which I feel should be
separated from the directory issue.


psteinb (approved):
LGTM

Issue #233, opened 2020-11-02
License file is outdated


dpshelio:
The information about the Carpentries in the [`LICENSE.md`](https://github.com/hpc-carpentry/hpc-intro/blob/358c5bb456ee384f30a7f378acd52e69bd093e87/LICENSE.md) file is outdated. Also, it would be good to make reference to what HPC-Carpentry is in respect with Software Carpentry (the logo used in the lessons) and the rest around it. E.g., is the right ["contact"](https://github.com/hpc-carpentry/hpc-intro/blob/358c5bb456ee384f30a7f378acd52e69bd093e87/_config.yml#L77) the carpentries itself?
This applies to all the lessons! (except the e-mail bit which is empty for `hpc-chapel`) So I'm not sure whether create issues in them all or use the coordination one. I've used this one as it's the one with more action 😁

PR #234, opened 2020-11-03
(this is only an exercise pull request) update license text

bast:
Part of #233.

This is only a demo during teaching.


rkdarst:
This is a sample reply.

Looks like a good start.

tkphd:
Superceded by #238.

PR #236, opened 2020-11-26
create file for learner profiles

psteinb:
- stub content taken from https://github.com/hpc-carpentry/hpc-carpentry.github.io/blob/master/why-hpc-carpentry.md#environmental-biology
- create file structure
- adapt the 'learner profiles' from lesson module 11

Closes #236


tkphd (approved):
LGTM!

PR #240, opened 2020-12-08
Sync with upstream style

### LINKED ISSUE #239: "make serve" forces global config change to bundle
Hi, I noticed the recent change in `JEKYLL` var in lesson's `Makefile` leads to a global change in my `~/.bundle/config` that is undesirable. I have MacOS with ruby fed from Homebrew. I have been installing all the Jekyll gem dependencies to build Carpentry-style webpages with "bundle install" once, and for all lessons I was able to simply run "bundle exec make serve". But the recent change to hpc-intro in commit 4b880e32b9304dc964ec557342cea8d9c0c0211f causes `.vendor/bundle` to be set for *every* lesson repo I have--causing fatal error every time I wanted to compile lessons other than hpc-intro. Turned out that this config would force me to do "bundle install" for every lesson repo I have (and I have many of them--over 10). @ocaisa , may you please explain why the change in the JEKYLL definition in the first place? I am trying to avoid installing lesson-specific gems for every lesson if at all possible.

ocaisa:
As per https://github.com/carpentries/styles/pull/473

Fixes #239

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -6,7 +6,7 @@ export SHELL = /bin/bash
 
 # Settings
 MAKEFILES=Makefile $(wildcard *.mk)
-JEKYLL=bundle config set path '.vendor/bundle' && bundle install && bundle update && bundle exec jekyll
+JEKYLL=bundle config --local set path .vendor/bundle && bundle install && bundle update && bundle exec jekyll
 HPC_JEKYLL_CONFIG?=
 PARSER=bin/markdown_ast.rb
 DST=_site


ocaisa:
@wirawan0 Can you verify that this solves the problem you were seeing?

tkphd (approved):
LGTM :+1:

wirawan0:
Great, thanks Alan and Trevor!

psteinb:
Merging this after @tkphd positively reviewed this.

PR #246, opened 2021-01-28
don't load modules in every shell

### LINKED ISSUE #61: Loading a module by default
14-modules talks about putting 'module load' commands in .bachrc and .bash_profile.

1. I generally discourage users from doing this. It often slows down their login session, they forget they have them and then later open tickets about things behaving oddly, module names can change, etc. 

2. If it's to be kept, we should at least explain the difference between the two conf files, and issue a word of caution about why it may cause problems

tkphd:
Hacking `.bashrc` is the wrong way to auto-load modules. Closes #61 by removing the last reference to that notion.


psteinb (approved):
Full in favor!! Well spotted.

PR #247, opened 2021-01-28
wget locally, then upload to firewalled cluster

### LINKED ISSUE #62: address internet connectivity issues common to clusters
15-transferring-files talks about using wget to grab files from the internets.  We should also include a warning about some clusters not being able to get to the internet at all, or maybe only a head node or DTN node being setup for internet access.

tkphd:
Closes #62.


psteinb (approved):
Ready to go.

PR #249, opened 2021-01-28
Note the subtle error when copy-pasting yields — instead of -- in flags

### LINKED ISSUE #70: "Customising a job" section, pay attention to copy past.
One regular issue we have in HPC at UC Merced is with copy past. 
Some OS/text editor/etc try to be "helpful" and convert `--`/`-` to em dashes, and [`u0096`](http://www.fileformat.info/info/unicode/char/96/index.htm) it can be relatively hard to diagnose, in the cli you can use `grep $'u0096'` (on bash 4.2+ I believe) to find those. 

It is relatively sneaky as there is no errors, and flags get just ignore on sbatch. 
This is rare but recurrent for user who create batch script on their machine in word/text-edit and then upload to the cluster. 

Might not be a good thing in the lesson itself, but for the instructors notes.

tkphd:
Appears under Extras -> Guide.

Closes #70.

### DIFF
--- a/_extras/guide.md
+++ b/_extras/guide.md
@@ -4,36 +4,59 @@ title: "Instructor Notes"
 permalink: /guide/
 ---
 
+## "No such file or directory" or symbol `0096` errors
+
+`scp` and `rsync` may throw a perplexing error about files that very much do exist. One source of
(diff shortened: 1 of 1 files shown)


reid-a (approved):
Looks good.

PR #252, opened 2021-01-28
explain that we won't run jobs on the login node in this lesson

### LINKED ISSUE #250: running jobs on login nodes
*This Issue was broken out of #79.*

In [12_cluster.md, line 140](https://github.com/carpentries-incubator/hpc-intro/blame/gh-pages/_episodes/12-cluster.md#L140), we have 

> A login node serves as an access point to the cluster. As a gateway, it is well suited for uploading and downloading files, setting up software, and running quick tests. It should never be used for doing actual work.

Opinions differ on this last point. Are we best served just deleting it?

- What constitutes "actual work?"
    - [ ] Compiling software
    - [ ] Visualizing results
    - [ ] Running debug cases
    - [ ] Running full code to get a sense of runtime
    - [ ] Running when the queue is saturated
    - [ ] Running when you forget how to `slurm`
- Is there a threshold cluster size below which running on the login node is all right? 
    - [ ] What if there's just a workstation, no dedicated login node?
    - [ ] One dual-homed login node & one compute node?
    - [ ] Heterogeneous layout, maybe a Raspberry Pi landing pad and x86 compute nodes? (Who would do such a thing???)
- Is there a threshold above which the login node must only be used for signing in?
    - [ ] Assume there's a login server *pool* at this scale?

tkphd:
Elegant solution to the question, provided by @reid-a. 

Closes #250.

### DIFF
--- a/_episodes/12-cluster.md
+++ b/_episodes/12-cluster.md
@@ -136,8 +136,13 @@ Individual computers that compose a cluster are typically called *nodes* (althou
 hear people call them *servers*, *computers* and *machines*). On a cluster, there are different
 types of nodes for different types of tasks. The node where you are right now is called the *head
 node*, *login node*, *landing pad*, or *submit node*. A login node serves as an access point to the
-cluster. As a gateway, it is well suited for uploading and downloading files, setting up software,
-and running quick tests. It should never be used for doing actual work.
+cluster.
+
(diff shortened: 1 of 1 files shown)


reid-a (approved):
Beat me to the PR. Approved.

PR #262, opened 2021-01-30
explain how to serve the customized site on GitHub

### LINKED ISSUE #196: Document how to deploy non-default site
While setting the environment variable `SITE_CONFIG` allows us to test builds for site other than Graham, deploying this through GitHub Pages will not see that variable.

The current pathway is the old workflow, overwriting `_config.yml` with the relevant content from `_config_settings.yml`. If there's a more elegant way to do it, e.g. some way to set an environment variable within Jekyll, that would be cool.

This needs to be documented somewhere, either `README.md` or `index.md`.

tkphd:
Requires templating a personal fork of the site and modifying config files.

Closes #196.


reid-a (approved):
Looks good.

PR #263, opened 2021-02-02
Moved lesson order.

### LINKED ISSUE #230: Introduce non-sequential lesson numbering
This issue is the successor of #226, which was partially addressed by renaming the snippet directories.

The remaining issue is that the lesson files under `_episodes` have sequential numeric prefixes, which controls the order in which they appear in the built website. Getting this order right is important, but the sequential numeric prefixes hinder flexibility.

However, moving away from this is not entirely trivial, because the build-test scripts assume the existence and sequential character of these prefixes. Any change to this involves a change to the build-tests, which came from the upstream Carpentries style, and so should be coordinated with them.

There is an [upstream issue](https://github.com/carpentries/styles/issues/502) on this.

Obvious options are to have gaps in the numeric ordering, or to just use descriptive names, and control the ordering [manually](https://github.com/carpentries/styles/blob/gh-pages/_includes/manual_episode_order.html).

reid-a:
Lesson order is now specified in the site-specific _config_options.yml file.  Updated the top-level README to reflect this. Did not actually re-order the lessons.

Closes #230.

### DIFF
--- a/_includes/snippets_library/ComputeCanada_Graham_slurm/_config_options.yml
+++ b/_includes/snippets_library/ComputeCanada_Graham_slurm/_config_options.yml
@@ -53,3 +53,13 @@ sched:
   info: "sinfo"
   comment: "#SBATCH"
   hist: "sacct -u yourUsername"
+
+episode_order:
+  - 11-hpc-intro
+  - 12-cluster
+  - 13-scheduler
+  - 14-modules
+  - 15-transferring-files
+  - 16-parallel
+  - 17-resources
+  - 18-responsibility
--- a/_includes/snippets_library/ComputeCanada_Graham_slurm/_config_options.yml
+++ b/_includes/snippets_library/ComputeCanada_Graham_slurm/_config_options.yml
@@ -63,3 +63,4 @@ episode_order:
   - 16-parallel
   - 17-resources
   - 18-responsibility
+
--- a/README.md
+++ b/README.md
@@ -30,7 +30,9 @@ NOTE: This is *not* Carpentries boilerplate! Please read carefully.
      snippet in episode 12 is included using `{% include /snippets/12/info.snip %}`.
 
 3. Edit `_config_options.yml` in your snippets folder. These options set such things as the address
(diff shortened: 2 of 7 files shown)


_includes/snippets_library/ComputeCanada_Graham_slurm/_config_options.yml
```diff
@@ -53,3 +53,13 @@ sched:
...
+
+episode_order:
+  - 11-hpc-intro
+  - 12-cluster
+  - 13-scheduler
+  - 14-modules
+  - 15-transferring-files
+  - 16-parallel
+  - 17-resources
+  - 18-responsibility
```
tkphd, line 65: Please add a newline. POSIX will thank you.

tkphd (changes requested):
Please terminate files with a newline character.

_includes/snippets_library/ComputeCanada_Graham_slurm/_config_options.yml
reid-a: Done.

tkphd (changes requested):
Looks pretty good! Thanks for propagating throughout the snippet library.
My feeling is that `episode_names` should also be defined in the top-level `_config.yml`. Do you have a reason not to place it there?

reid-a:
The top-level behavior is to run through the entire contents of _episodes in lexical order, I think, which is a reasonable default, but making it explicit probably improves legibility. The downside is that it's yet another place where the file-names are hard-coded.

On balance, I favor this, and have now done it.

tkphd:
Yeah... I prefer having the episode order defined explicitly at the top, then we can document changes. When the site builds on GitHub, it only uses what's in `_config.yml`, but local copies -- and those on GitLab, which provides a config file for the build system -- can use HPC_JEKYLL_CONFIG to override the top-level settings. If a site's snippet config defines a different order (like CTCMS dropping the modules lesson), those local changes are kept. Otherwise, default; or, as is currently the case, every snippet lib also explicitly defines lesson order.

tkphd (approved):

PR #267, opened 2021-02-03
Improve program to estimate π.

### LINKED ISSUE #266: Pi computes the same numbers on each rank
Since the PRNG is seeded with a constant, each MPI rank has the same state, and therefore generates the same sequence of numbers.

tkphd:
- Each rank gets its own random seed. Fixes #266.
- L10n for , & . grouping separators.
- State that the code runs on [0, 1) not (-1,1).
- Pythonic print formatting (no hard-coded types).

### DIFF
--- a/files/pi.py
+++ b/files/pi.py
@@ -1,49 +1,60 @@
 #!/usr/bin/env python3
 
-"""Parallel example code for estimating the value of pi.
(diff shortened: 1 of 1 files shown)


reid-a (approved):
Looks good.

Issue #268, opened 2021-02-05
live code parallelization of pi.py


tkphd:
Could we start with a serial version of [pi.py](https://github.com/carpentries-incubator/hpc-intro/blob/gh-pages/files/pi.py), and have learners live-code the upgrade to MPI? As far as I can tell, this would involve
1. Importing mpi4py
2. Setting the MPI communicator, then getting its size & the local rank
3. Dividing the number of points evenly between processors (or not -- we could have each processor compute the specified number?)
4. Performing the MPI Gather, then locally sum; or MPI Reduce
5. Introducing a conditional on the output, so that only rank 0 reports the final value.

It could take 15 minutes to teach through it, but I don't think it's an unreasonable level of detail, esp. since Python abstracts away a lot of boilerplate and leaves just the interesting pieces exposed.

tkphd:
As written, `pi.py` could be used for an Amdahl's Law scaling study. If we simplify the code so that each rank generates the number of points specified, we can do a Gustafson's Law scaling study. Either way, we win!

PR #269, opened 2021-02-05
Sprawling curriculum updates

### LINKED ISSUE #244: adapt Nelle Nemo exercise for clusters
As mentioned in #21 by @pdoehle, ["Not So Fast Nelle: Taking Advantage of Parallelism"](https://github.com/OSU-HPCC/hpc-lessons/blob/master/05_nelle_parallel.md) brings parallel processing to existing Carpentries material, demonstrating a parallel workflow with clarity. It uses GNU parallel to do this; for a cluster environment, the queuing system is a more appropriate tool. Let's port the Nelle Nemo exercise as the capstone for hpc-intro, using a reconfigurable job script, job arrays, or similar.

### LINKED ISSUE #255: purge fastqc
`fastqc` is currently included as a relatively simple program users can copy from GitHub, compile, and just run as an exercise for using the queuing system.

Pedagogically, this exercise is flawed: we're leading learners through a set of instructions without the need to understand what it's for or why it's useful to do it. It would be preferable to devote that time to teaching what MPI is and how message passing works for distributing workloads across nodes, introduce & explain the `pi.py` script, and run that in parallel using increasing numbers of cores.

tkphd:
This PR represents combined work overhauling the curriculum to present parallel concepts more clearly, get rid of some needless complications, and reflect some changes to improve pedagogy in teaching core concepts.

- Updated the lesson outlines in README with updated file names, lesson titles, and runtime estimates. We're at 5&frac12; hours, if the numbers have any basis in reality.
- Purged `fastqc` and its dependent files, since explaining what it does & why is more effort than its worth.
- Created an MPI-parallelized `goostats.py` to work on Nelle Nemo's data, using MPI Reduce.
- Wrote, then reverted, Nelle Nemo-based material.
- Lightly reworked Ep. 17 to use the MPI-parallelized `pi.py` code to look into runtime stats.

Upside: the lessons are lighter-weight and more conceptually consistent and clear.

Downside: the discussion of downloading & unpacking a tarball in Ep. 15 is rendered moot, since those files are never mentioned again.

Work remaining: the snippet libraries for Ep. 17 do not yet contain runtime analysis of the `parallel-pi` job. A more thorough revision of that episode to introduce a scaling study would be warranted (#268). 

Closes #244. 
Closes #255.


reid-a (approved):
This is a solid and substantive improvement. There are minor editorial issues that can easily be taken up in a separate issue or isssues.

tkphd:
Thanks, Andrew. I'll hold off merging until at least Monday, so @mikerenfro and others have a chance to finish reviewing.

ChristinaLK:
I'm tagged, but don't have the bandwidth to review - sorry!

tkphd:
Rebased; preparing to merge.

README.md
```diff
@@ -120,30 +120,27 @@ converted into a flow of learning goals and questions. Note, again, that it is p
 building your actual lesson, to re-order these files, or omit one or more of them.
 
 * Prelude
-    * User profiles (academic and/or commercial) of cluster users
+    * [User profiles](
+      https://github.com/carpentries-incubator/hpc-intro/blob/gh-pages/_extras/learner-profiles.md)
```
ocaisa, line 124: This link was added but the corresponding file was also removed in the same PR. I'll open a PR to reinstate the file and we can decide exactly what to do with it. Maybe this link just needs an update?

tkphd: Oops :laughing:

tkphd: I think we should update the link, rather than reinstating the file.

ocaisa: There's no real good way of doing that because of the auto-rendering of markdown, best I can do is https://github.com/carpentries-incubator/hpc-intro/blame/gh-pages/_episodes/11-hpc-intro.md#L23

I'll open a PR that adds a heading to that section so I can link to that instead

tkphd: A couple of options:

1. Don't make it a direct link, as the profiles are the first thing in the lesson (with no lead-in... something else to fix).
2. Refer to the Learner Prerequisites (#308) instead of the lesson content, as that is actually "prelude" material.
2. Drop the "prelude" item, and move the mention of learner profiles (without linking) to the next list item, where it now belongs.

PR #272, opened 2021-02-07
Update 11-hpc-intro.md

mikerenfro:
Introducing concepts of serial and parallel earlier, expanding on why each research problem is suitable for HPC.


tkphd:
I like it! For flow, instead of calling back, perhaps we could combine the detail on each use case into the enumerated list, or dissolve the bulleted list and merge them with your detailed exposition.

mikerenfro:
> I like it! For flow, instead of calling back, perhaps we could combine the detail on each use case into the enumerated list, or dissolve the bulleted list and merge them with your detailed exposition.

Went with the first option for now.

tkphd (approved):
This is great!

PR #275, opened 2021-02-08
WIP: Pi calculations

mikerenfro:
Some additional materials related to testing Python MPI code, and related to #268. General review requested, and I'll wait until the sprawling curriculum updates have been merged before touching anything outside this folder.

Baseline serial data showing scaling with respect to problem size can be generated with something like:

```
for n in 1 2 4 8 16 32 64 128 256 512 ; do ./pi-serial.py ${n}000000; done > pi-serial.txt
```

where the values of `n` should get large enough to not run out of memory on the designated computer, but large enough to take several seconds to complete.

After the largest problem size for the serial case has been determined, the MPI performance data can be generated with something like:

```
for n in 1 2 4 8 16 28 56 84 112; do mpirun -np ${n} ./pi-mpi.py 512000000 | tail -n 1; done > pi-mpi.txt
```

These files can then be compared with: `./pi-plot-parallel-results.py pi-mpi.txt pi-serial.txt`, which will generate both elapsed time graphs and a more interesting speedup graph as attached.

HPC, with up to 4 28-core nodes. shows relatively consistent speedup through 3 nodes:
mpi_Speedup_factor

8-core Laptop, shows the expected diminishing return once all 8 cores are in use:
laptop-mpi_Speedup_factor

Initial conclusions that might be useful to include in lessons:

- both graphs show considerable overhead for MPI on small core counts
- some newer laptops may be faster than some older HPCs for small core counts


reid-a:
Thanks Mike, this is important to convey!

In the second figure, the first data point for one core is below "1" on the Y-axis, "speed-up factor". Is there a reason for this? If I had to guess, I'd guess you're actually measuring speed, and there's a secret other Y-axis that's not in the figure?  But for the purpose of the narrative we're trying to convey, it would be good if these were aligned. (Maybe this is fixed in the subsequent commits below the figure?)

mikerenfro:
> Thanks Mike, this is important to convey!
> 
> In the second figure, the first data point for one core is below "1" on the Y-axis, "speed-up factor". Is there a reason for this? If I had to guess, I'd guess you're actually measuring speed, and there's a secret other Y-axis that's not in the figure? But for the purpose of the narrative we're trying to convey, it would be good if these were aligned. (Maybe this is fixed in the subsequent commits below the figure?)

Honestly, the 2-core runs are probably under 1.0 on the y axis as well. I assume this is due to MPI setup and teardown overhead. But I'll take a second look and make sure my speedups aren't miscalculated (they should be using the serial version of the Pi calculator for the given problem size as the baseline time).

mikerenfro:
Found an off-by-one on finding my baseline time, so no real MPI overhead as I'd first thought.

Laptop:
laptop-mpi_Speedup_factor

HPC:
hpc-mpi_Speedup_factor

tkphd:
Please use the `locale` package to format the numbers with grouping. The general idea is 

```python
import locale
locale.setlocale(l10n.LC_ALL, "")

...

plt.title("MPI π results with {:n} samples".format(512000000000))
```

This should result in the title being a more legible **MPI π results with 512,000,000,000 samples**.
If the library works the way I hope, then for those preferring '.' as their grouping delimiter, it will automatically generate "MPI π results with 512.000.000.000 samples" instead. Note that, since GitHub is 'murican, it will always use ','.

tkphd (changes requested):
Please place images under /fig instead of /files, and create entries for them in _includes/all_figures.html.

tkphd:
"allone" variable name triggers a spellchecker error. Consider underscores or camel case to try clearing it.

mikerenfro:
> "allone" variable name triggers a spellchecker error. Consider underscores or camel case to try clearing it.

Can't underscore a macro name, but in modifying the code to match the Python variable names, `all` is no longer relevant.

Initialize:
initialize

Scatter 1:
scatter-1

Scatter 2:
scatter-2

Compute:
compute

Gather:
gather

Finalize:
finalize

tkphd (approved):

files/pi-mpi.py
```diff
@@ -0,0 +1,125 @@
...
+        elapsed_time = (datetime.datetime.now() - start_time).total_seconds()
+
+        # Memory required is dominated by the size of x, y, and radii from
+        # inside_circle(), calculated in MiB
+        size_of_float = np.dtype(np.float32).itemsize
+        memory_required = 3 * n_samples * size_of_float / (1024**2)
+
+        # accuracy is calculated as a percent difference from a known estimate
+        # of π.
+        pi_specific = np.float32(3.141592653589793238462643383279502884197169)
```
tkphd, line 118: float32 truncates this to `3.1415927410125732421875`.
Use `np.pi` instead.

files/pi-mpi-hpc.txt
```diff
@@ -0,0 +1,9 @@
+1,512000000,5859.375000,28.546736,0.003635
```
tkphd, line 1: - These `pi-*.txt` files represent CSV data. Please change their extensions and add header rows.
- Are these files to be used in the lesson? If not, consider bundling the results, and plot generator scripts, in a tarball.

tkphd (changes requested):

tkphd:
Ack, I meant to approve the changes requested, only, not the entire PR. Oops!

PR #277, opened 2021-02-08
Added caution notes about recursive copying

pablobernabeu:
Hello,

Thank you for the tutorial.

I've suggested adding caution notes about about recursive `-r` copying.

New users may be overwhelmed if they turn to the recursive option--for instance, because the standard `scp` isn't working--and see the files being multiplied non-stop.


psteinb (approved):
LGTM. Thanks for the contribution.

PR #278, opened 2021-02-08
rephrase `sleep 60` example to avoid exclusive node access

### LINKED ISSUE #159: using one full node
In Ep. 13 "Submitting resource requests," poses this challenge: "Submit a job that will use 1 full node and 1 minute of walltime."

To "use 1 full node" would require the `--exclusive` flag in Slurm, or its equivalent. However, I suspect this is not what is meant. Rather, we want to use *no more than 1 node*, and 1 minute of walltime. This would be consistent with the rest of the text, which leaves the defaults (one node, one CPU, or `-N1 -n1`) alone. 

Revision for better clarity (and accuracy in the challenge) would be welcome.

tkphd:
Closes #159.

### DIFF
--- a/_episodes/13-scheduler.md
+++ b/_episodes/13-scheduler.md
@@ -262,7 +262,8 @@ about how to make sure that you're using resources effectively in a later episod
 
 > ## Submitting resource requests
 >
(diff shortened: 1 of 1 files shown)


reid-a (approved):
Looks good.

PR #285, opened 2021-02-08
Rephrase scatter and gather summaries for correctness and clarity.

### LINKED ISSUE #279: More precise wording on scatter/gather
The recently-updated parallel lesson has descriptive text which is possibly misleading, it defines:

 * Scatter: an MPI operation wherein the process running on one rank sends a piece of 
     information (value, array, or object) to all others on the given comm.
 * Gather: the inverse of Scatter, one rank receives a piece of information from all

This is imprecise.

What scatter does more precisely is partition an array and send different array fragments to the partner ranks, with each rank getting a different portion of the array. The current text is a description of a broadcast operation, where all of the partner ranks get the same data from the sending rank.

What gather does more precisely is populate different portions of a local array with data from the various partner ranks. There maybe many "pieces of information".

Having gone through the MPI documentation, i can find no indication that scatter and gather can work on non-array objects, but I am not an MPI expert.

A description of "scatter" that overlaps with what "broadcast" actually does is potentially confusing for newcomers to the concept of parallelism and data heterogeneity.

Proposed revision:
* Scatter: An MPI operation in which an array of data is divided up, with separate portions being sent out
    to the partner ranks. Each partner rank receives data from a different part of the array.
* Gather: The inverse of scatter. One rank populates a local array, with different portions of the array populated
    by data from different partner ranks.

tkphd:
Closes #279.

### DIFF
--- a/_episodes/16-parallel.md
+++ b/_episodes/16-parallel.md
@@ -80,14 +80,18 @@ clearly. Press "q" to exit the pager program (`less`).
 >
 > This relatively simple program exercises four important concepts:
 >
-> * COMM_WORLD: the default MPI Communicator, providing a channel for all the processes
->   involved in this `mpirun` to exchange information with one another.
-> * Scatter: an MPI operation wherein the process running on one rank sends a piece of
->   information (value, array, or object) to all others on the given comm.
-> * Gather: the inverse of Scatter, one rank receives a piece of information from all
->   others on the given comm.
-> * Conditional Output: since every rank is running the *same code*, the general `print`
->   statements are wrapped in a conditional so that only one rank does it.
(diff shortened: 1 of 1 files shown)


psteinb (approved):
LGTM. Some eye candy would be great to bring this into a picture to grasp easily.

mikerenfro:
Minimum viable graphics below. Can provide Inkscape file on request.

Scatter:
scatter

Gather:
gather

tkphd:
Thanks @mikerenfro. This is illustration is at odds with the lesson text: there must be one, and only one, array element per MPI rank. It would also be helpful to have values that are not easily mistaken for rank indices in the array. Please open a separate issue, as this PR is Closed.

Issue #221, opened 2020-08-14
Visualize bandwidth vs. latency in data transfers


tkphd:
In Ep. 18, possibly a general "bandwidth vs latency" paragraph in the
context of scaling would be useful? Amdahl's law, even?

tkphd:
Spun out from #185.

mikerenfro:
Looks like this is Episode 18 now. Just to save others the confusion I had.

reid-a:
Another reason to use names to describe lessons, and not sequence numbers.

mikerenfro:
In the broader context of bandwidth and latency for disk and network I/O, I've got a figure attached that may help visualize things: [disk-network-bandwidth-latency.pdf](https://github.com/hpc-carpentry/hpc-intro/files/5435825/disk-network-bandwidth-latency.pdf). Each of the components on the figure is connected by a black line of width proportional to the interface bandwidth, and a white line of width proportional to the data transferred along that interface.


```latex
\documentclass{standalone}
\usepackage{amsmath}
\usepackage{tikz}
\usetikzlibrary{positioning}
\begin{document}
\newcommand\dataline[5]{%
\draw [line width=#1 mm] (#3) to [#5] (#4);%
\draw [line width=#2 mm,white] (#3) to [#5] (#4);%
}
\begin{tikzpicture}%[]
[disk/.style={rectangle,draw=black!50,fill=black!20,align=left},
 controller/.style={rectangle,draw=red!80,fill=red!20,align=left},
 nic/.style={rectangle,draw=blue!80,fill=blue!20,align=left},
 switch/.style={rectangle,draw=green!80,fill=green!20,align=left}]
\node[disk]       (disk 1)                                     {SSD 1, 12 Gb/s interface,\\ sends 4 Gb/s bandwidth,\\ 4 ms latency};
\node[disk]       (disk 2)          [below=of disk 1]          {SSD 2, 12 Gb/s interface,\\ sends 4 Gb/s bandwidth,\\ 4 ms latency};
\node             (dots)            [below=of disk 2]          {$\vdots$};
\node[disk]       (disk 9)          [below=of dots]            {SSD 9, 12 Gb/s interface,\\ sends 4 Gb/s bandwidth,\\ 4 ms latency};
\node[disk]       (disk 10)         [below=of disk 9]          {SSD 10, 12 Gb/s interface,\\ sends 4 Gb/s bandwidth,\\ 4 ms latency};
[...]

PR #287, opened 2021-02-08
visualize bandwidth vs latency in data transfers

tkphd:
Incorporate @mikerenfro's bandwidth diagram as SVG.
Addresses #221, though text will require further updates.


reid-a (changes requested):
This is a solid start.

There are some issues, in descending order of importance.

The major issue I have is that it's not immediately obvious from the diagram that the left-most three columns are all on a single storage server. This is not relevant to the data path, but since the lesson narrative has been centered on nodes, it would be useful to clarify where the node boundaries are in this diagram.

An important but difficult (for me) issue is that the diagram doesn't convey latency. A major caveat is that I'm not sure how one would do this, but conspicuously absent from the figure is the fact that the RAID controller can hide latency by pipelining request operations to the individual drives, so the effective latency of the storage path is actually lower than latency of the worst-performing link in the path. This may be too subtle a point for the lesson, the broader point of avoiding many small transactions with the storage system is of course a good one.

Since the diagram is schematic, it's not obvious why the links connecting the rightmost three columns are twinned.

tkphd:
Pipelining is too granular. Perhaps latency could be encoded in length of the wires?
The CPU is also missing, I think PCI bus devices can't talk directly to one another. Might be wrong.

reid-a:
So having looked at it again, I now understand that the right-most three columns are not "twinned", that's a white line inside a black line. Possibly the data-transfer should not be the same color as the background, so that other people don't make the same mistake?

I can certainly live with leaving out the discussion of pipelining.

Possibly latency could be indicated by a gap between the line and the box, at the point where they would otherwise meet? The model is, the data needs to hop over the gap, and then it's clear sailing down the line.

tkphd:
Iteration:

responsibility-bandwidth

```tex
\documentclass{standalone}
\usepackage{amsmath}
\usepackage{tikz}
\usetikzlibrary{arrows, calc, backgrounds, positioning}

\newcommand\dataline[4]{
    \draw [line width=#1 mm] (#2) to [#4] (#3);
}

\tikzstyle{background} = [fill=green!5!white,draw=green!75!black,very thick]
\tikzstyle{titlebox} = [fill=green!75!black,text=black,rounded corners]

\tikzstyle{disk}       = [rectangle,draw=black!50,fill=black!20,align=left,rotate=90]
\tikzstyle{controller} = [rectangle,draw=red!80,fill=red!20,align=left]
\tikzstyle{network}    = [rectangle,draw=blue!80,fill=blue!20,align=left]
\tikzstyle{switch}     = [rectangle,draw=green!80,fill=green!20,align=left]

% Network latency is 0.004 ms
%  Disk   latency is 4     ms

% If the sep between NIC and switch is 2cm, then using a log scale, the
% disks are (2 cm) * log(4 / 0.004, 2) = 13.8 cm below the RAID controller

\begin{document}
    \begin{tikzpicture}
        \node[controller]         (raid)                                           {RAID controller, 63 Gb/s interface,\\ forwards 40 Gb/s bandwidth};
        \node[disk,anchor=north]  (disk 1)     at ($(raid.south)+(-4cm, -13.8cm)$) {SSD 1, 12 Gb/s interface,\\ sends 4 Gb/s bandwidth,\\ 4 ms latency};
        \node[disk]               (disk 2)     at ($(disk 1) + (2cm, 0)$)          {SSD 2, 12 Gb/s interface,\\ sends 4 Gb/s bandwidth,\\ 4 ms latency};
        \node                     (dots)       at ($(disk 2) + (1.5cm, 0)$)        {$\cdots$};
        \node[disk]               (disk N)     at ($(dots) + (+1.5cm, 0)$)         {SSD $N$, 12 Gb/s interface,\\ sends 4 Gb/s bandwidth,\\ 4 ms latency};
        \node[network]            (nic)        [right=of raid]                     {network card, 56 Gb/s interface,\\ forwards 40 Gb/s bandwidth};
        \node[switch,anchor=west] (switch)     at ($(nic.east) + (2cm, 0)$)        {network switch, 56 Gb/s interfaces,\\ forwards 40 Gb/s bandwidth,\\ 0.004 ms latency};
        \node[network]            (node 1 nic) [above=of switch]                   {node 1 network card, 56 Gb/s interface,\\ receives 20 Gb/s bandwidth};
        \node[network]            (node 2 nic) [below=of switch]                   {node 2 network card, 56 Gb/s interface,\\ receives 20 Gb/s bandwidth};

        \draw[line width=1.2mm, out=90, in=260] (disk 1.east) to node[above,rotate=70] {4 ms} (raid);
        \dataline{1.2}{disk 2.east}{raid}{out=90,in=270}
        \dataline{1.2}{disk N.east}{raid}{out=90,in=280}
        \dataline{6.3}{raid}{nic}{}
        \draw[line width=5.6mm] (nic) to node[above]{4 ns} (switch);
        \dataline{5.6}{switch}{node 1 nic}{}
        \dataline{5.6}{switch}{node 2 nic}{}

        \node[titlebox] (server) at ($(raid.north)+(0cm, 1cm)$) {\textbf{Server}};

        \begin{pgfonlayer}{background}
                    \draw [background] ($(server.west)+(-4cm, 0)$) rectangle ($(nic.east)+(0cm, -16.8cm)$);
        \end{pgfonlayer}
    \end{tikzpicture}
\end{document}
```

reid-a:
This is a definite improvement, and I definitely appreciate the effort. 

I'm still uncomfortable with how latency is depicted, and still not completely sure what to do about it. In the revised diagram, it looks like "latency" means "transit time", which is a *kind* of latency, in that it's annoying in a way that's independent of bandwidth. But the kind of latency that's relevant to storage systems is the kind where it takes a certain amount of time to set up the transmission -- to seek to the relevant disk sector, or assemble the packet and put it on the wire, or shunt all the bits in order out of the flash chips. It's a per-transaction cost that doesn't go away when you increase bandwidth. This only fits the revised figure if you make a further assumption that only one transaction can be on the link at a time, which is not true in general.

I'm still partial to little gaps, or maybe resistors or something, at the point where the lines meet the boxes.

tkphd:
Little gaps or resistors make less sense to me, either suggesting wireless hard drives or introducing schematics that are hard to grasp, and hard to make proportional to the actual latency. Your point is well taken, though, as the latency is not primarily due to the length of the wires (at *c*, signals traverse those pretty quickly). What about a little maze graphic at the interface between device and wire, with complexity proportional to the latency?

reid-a:
Little maze diagrams would work reasonably well, they associate the effect with the transition to the link and not the link itself, which is desirable. I don't think strict proportionality is important, just that the SSDs have visibly more latency than the controller, but if you can manage proportionality, I am not opposed.

tkphd:
Like so?

responsibility-bandwidth

tkphd:
Well, that's some ugly conversion artifacting. Take a look at the [SVG](https://github.com/tkphd/hpc-intro/blob/issue221-visualize_bandwidth_vs_latency_in_data_transfers/fig/responsibility-bandwidth.svg).

mikerenfro (approved):
Looks reasonable to me.

reid-a:
Looks OK to me. I'd prefer if the mazes were smaller, but at this point we're into increasingly minor aesthetic preferences.

tkphd:
"Smaller," in what way? Less-complicated mazes, or smaller thumbnails, or... ?
Mazes were generated with [pymaze](https://github.com/jostbr/pymaze) and the following script; pathways were added in GIMP.

```python
from __future__ import absolute_import
from src.maze_manager import MazeManager
from src.maze import Maze


if __name__ == "__main__":

    # Set maze size
    W = 5 * 12
    H = 5 * 8

    # The easiest way to use the library is through the Manager class. It acts as the glue between
    # The visualization, solver, and maze classes. Mazes inside the manager have unique ids that we use
    # to specify particular mazes.
    manager = MazeManager()

    # If we want to save the maze & maze solution images along with their animations, we need to let the manager know.
    manager.set_filename("medium")

    # We can add mazes to the manager two different ways.
    # The first way, we specify the maze dimensions. The maze that is created gets returned back to you.
    maze = Maze(H, W, start = (0, int(W/2)), finish = (H-1, int(W/2)))
    maze = manager.add_existing_maze(maze)

    # Show the generated maze
    manager.show_maze(maze.id)

    # Once we have a maze in the manager, we can tell the manager to solve it with a particular algorithm.
    # manager.solve_maze(maze.id, "BreadthFirst")
    # manager.solve_maze(maze.id, "BiDirectional")
    manager.solve_maze(maze.id, "DepthFirstBacktracker")

    # Finally, you can show an image of the maze with the solution path overlaid. All of these display
    # functions will save the figure if MazeManager::set_filename has been set.
    manager.show_solution(maze.id)
```

reid-a:
The general aesthetic problem (for me) is that the "dynamic range" of feature sizes in the image is too wide, the principal culprit being the high-complexity mazes adjacent to the SSDs conveying (correctly!) their high latency. In order to be legible at all, these complex mazes need to be geometrically large, and even so have fine lines which, to my aesthetic sense, are both (importantly) extraneous detail, and (less importantly) the origin of the rendering artifacts in the preview. This hearkens back to my previous comment about proportionality of latency not actually being a requirement, in my opinion -- the important point to convey is that the SSD latency is larger than the controller latency.

Where I would go with this is try to come up with *icons* -- a low-complexity maze icon, and a high-complexity maze icon, not necessarily actual mazes. The icons would have feature sizes consistent with the rest of the diagram, and be sufficiently compact that they don't dominate the SSD figures, in particular. This would bring them into line with the schematic character of the rest of the figure, where e.g. the various components are simple labelled boxes, with no graphical reflection of their internal complexity.

It's not a simple balance, but the current version fails, to my eye, to be a legible abstraction of latency, although it's certainly an improvement on prior versions.

tkphd:
I'm failing at peeking inside your brain, but am happy to keep iterating -- confusion to your eye means it's failing to convey the concept clearly. Can you sketch something on paper & send? I'm confused about the difference between icons and the thumbnails currently presented.

tkphd:
Converted to a Draft so that this WIP doesn't get merged prematurely.

reid-a:
Sketch attached.  The basic idea is to have paths of different complexity through the same space, the space being a 4x4 grid of cells with barriers making it a maze. The simplicity of the space promotes legibility, and the feature-size of both images is the same, so they work together in an image. They differ in the complexity of the problem they are solving, but the problem is fundamentally the same, which is (hopefully!) reflected in the commonality of the space and the entry and exit points, and the difference in the arrangement of the barriers and the consequent complexity (number of turns, length of path) of the solution.

This sacrifices proportionality -- the complex maze is only a little more complicated than the simple one, but it conveys the idea of differing complexities of what is essentially the same task, which is getting data out of the device and into the transfer channel.

Disclaimer: Pretty sure Enthought doesn't endorse any of this.

PXL_20210211_153626664_e

tkphd:
Ah-ha, tortuosity vs size. Will try to address, thanks Enthought!

tkphd:
Better?

responsibility-bandwidth

reid-a:
Probably they could be a bit simpler still, for legibility, but at this point we are really into the minor-preference weeds.

tkphd:
You drive a hard bargain, mister! Final offer:

responsibility-bandwidth

reid-a:
Make it so.

reid-a (approved):
Thanks for sticking with it. Looks good now.

tkphd:
Thank *you* for the honest feedback! The lessons are better for it.

mikerenfro:
Late to the game, but a cleaner way to express tortuosity might be something like a helix structure with more loops indicating more latency. Shaped something like an inductor (via Wikipedia):
InductorSignalFilter1
But I'm good with any of this.

PR #289, opened 2021-02-10
Fix 288

### LINKED ISSUE #288: pi scaling visualization tweaks
The merged PR #275 contains some dangling issues.
- [x] Defining `pi_specific = np.float32(3.141592653589793238462643383279502884197169)` is misleading: single precision truncates this to `3.1415927410125732421875`. The built-in variable, `np.pi`, is vastly preferable, as it doesn't require us to define a new variable to compare, and it is a very high-precision representation.
- [x] Why are we even using float32? Upgrade to float64 throughout.
- [x] TXT files contain CSV data. Add header rows, and change extensions to `.csv`

mikerenfro:
Fixes #288, also archives the .py and .csv files into a tarball since they're not part of the lessons themselves.


tkphd (approved):
LGTM! :+1:

PR #294, opened 2021-02-12
Delete CNAME

### LINKED ISSUE #284: Deployment errors on non-canonical sites due to CNAME change?
When pushing recent changes up to my local fork of the repo, I got a message from github that the CNAME I requested, "hpc-intro.hpc-carpentry.org", is already taken, and not available to me.

I was directed to a [trouble-shooting page](https://docs.github.com/en/github/working-with-github-pages/troubleshooting-custom-domains-and-github-pages#cname-errors), which describes the issue in more detail.

Operationally speaking, this is a benign error, I don't really want my page to render on that URL, so I am not bothered by its failure to do so.

It's unclear to me how to avoid this. As far as I can tell, the CNAME file has to be part of the repo in order for the custom URL mechanism to work at all, and as part of the repo, it has to be reproduced in github forks of the repo, where it is doomed to fail. Thus, it seems that all the github forks are going to be broken, which doesn't seem right either.

ocaisa:
Trying to address #284 by using a CNAME record instead

### DIFF
--- a/CNAME
+++ b/CNAME
@@ -1,1 +1,0 @@
-hpc-intro.hpc-carpentry.org


ocaisa:
That was not going to work (see https://github.com/carpentries-incubator/hpc-intro/issues/284#issuecomment-778131422) but in my opinion we should still remove the CNAME record due to #284

reid-a (approved):
I think you're right, this is the way forward.

tkphd:
I'm not sure this is the way forward. 

Keeping the CNAME file lets us serve a consistent URL as we go through transitions.
Right now, #284 only applies to those of us actively maintaining our own forks: deployment only occurs on a push to the `gh-pages` branch, which is rare, and also a deliberate action. The resulting email from GitHub won't be a surprise, and can easily be ignored.

If we'd rather "go with the flow" and change the URL when the time comes, that's also fine, but I wanted to point out that we have options.

reid-a:
If we want a consistent URL we can publish that will always work, we can redirect from it, since "we" own the hpc-carpentry.org domain. I saw this suggested somewhere, I think from @ocaisa, but I'm not seeing it now.

tkphd (approved):
LGTM :+1:

Issue #286, opened 2021-02-08
visual representations of scatter and gather


tkphd:
From @psteinb's [comment](https://github.com/carpentries-incubator/hpc-intro/pull/285#pullrequestreview-585787669) on #285; see @mikerenfro's [concept diagram](https://github.com/carpentries-incubator/hpc-intro/pull/285#issuecomment-775365286).
Sketch of a step-by-step data diagram:
![scatter-sketch](https://user-images.githubusercontent.com/2179347/107269807-20fad180-6a18-11eb-82f7-ecaee93b2cb5.png)

tkphd:
Clarification: the interface exposed by mpi4py only allows one element per MPI rank. The more complete C and Fortran API includes integer values for the number of elements to send each rank (same number for each). There also exist MPI Scatterv and Gatherv, which allow you to send differing numbers of elements to each rank.

In the interests of simplicity and compatibility with existing Carpentries, I suggest focusing on the mpi4py interface, perhaps with a notice to that effect.

mikerenfro:
Could use a bit of highlight color, and possibly some refactoring of the TikZ styles, but:

![scatter-gather](https://user-images.githubusercontent.com/1451881/107392279-1f381900-6abf-11eb-98e8-cd1ecb47c006.png)

```latex
\documentclass{standalone}
\usepackage{tikz}
\usetikzlibrary{positioning}

\begin{document}
\tikzset{
  rank/.style={
    circle,
    very thick,
    draw=black
    },
  array/.style={
    rectangle,
  },
  node distance=25mm,
  arrow/.style={
    >=latex
  },
  line/.style={
    draw, ->
  },
}
\begin{tikzpicture}
  \node (rank0) [rank] {Rank 0};
  \node (rank2) [rank, below of=rank0] {Rank 2};
  \node (rank1) [rank, left of=rank2] {Rank 1};
  \node (rank3) [rank, right of=rank2] {Rank 3};
  \node (vars0) [above of=rank0, yshift=-18mm, label={[align=center]all = \([1.4, 2.3, 3.2, 4.1]\) \\ mine = 0}] {};
  \node (vars1) [below of=rank1, yshift=6mm, label={[align=center]all = \([]\) \\ mine = 0}] {};
  \node (vars2) [below of=rank2, yshift=6mm, label={[align=center]all = \([]\) \\ mine = 0}] {};
[...]

PR #296, opened 2021-02-14
Pi serial mpi episode

mikerenfro:
Let the review begin.


mikerenfro:
Tried to make the CI pipeline work with my Python code blocks, no luck yet. Might be an actual bug in bin/lesson_check.py.

tkphd (changes requested):
Code block tags use single braces -- `{: .bash}` and `{: .language-python}`, not `{{: .python }}`, etc.

mikerenfro:
> Code block tags use single braces -- `{: .bash}` and `{: .language-python}`, not `{{: .python }}`, etc.

Indeed they do.

tkphd:
Is this PR still a Draft? Please hit "Ready for Review" when it is so.

mikerenfro:
May also close #286 when merged.

tkphd:
Fellow reviewers (@reid-a esp), if you plan to build this branch locally, please check out the parent commit first. To wit:

```bash
you@yours:hpc-intro$ git checkout 81f90c6
Note: checking out '81f90c6'.

You are in 'detached HEAD' state. You can look around, make experimental
changes and commit them, and you can discard any commits you make in this
state without impacting any branches by performing another checkout.

If you want to create a new branch to retain commits you create, you may
do so (now or later) by using -b with the checkout command again. Example:

  git checkout -b <new-branch-name>

HEAD is now at 81f90c6 Merge pull request #290 from carpentries-incubator/fix-scatter-1-figure

you@yours:hpc-intro$ git checkout -b pi-serial-mpi-episode
Switched to a new branch 'pi-serial-mpi-episode'

you@yours:hpc-intro$ git pull origin pi-serial-mpi-episode
From https://github.com/carpentries-incubator/hpc-intro
 * branch            pi-serial-mpi-episode -> FETCH_HEAD
Updating 81f90c6..5888fcc
Fast-forward
«snip»

you@yours:hpc-intro$ make serve # &c.
```

reid-a:
Unclear to me why this isn't equivalent to just checking out the pi-serial-mpi-episode branch directly?

In both cases, I get a build-time error:

` [2021-02-18 10:12:35] ERROR `/fig/scatter.png' not found.`

Inspecting the file tree confirms this file is absent, but  `/fig/scatter-1.png` exists.

The fact that the CI checks all passed makes me think I am maybe misconfigured?

reid-a:
There are a couple of minor items not related to my infrastructure.

The legend in the performance figures identifies the meaning of the colored lines, but not the blue squares (which are obviously the observed timings, but it should still be described).

At the beginning, the conditional code block below `if __name__ == "__main__"` is described as a "main function". Strictly speaking, it's not a function, although it works like `main()` in C/C++. Maybe describe this as a top-level code block?

tkphd:
Hmm, good point. Was expecting it to do a silly merge commit, but it doesn't, so I need more coffee.

mikerenfro:
> The legend in the performance figures identifies the meaning of the colored lines, but not the blue squares (which are obviously the observed timings, but it should still be described).

Will get that edited.

> At the beginning, the conditional code block below `if __name__ == "__main__"` is described as a "main function". Strictly speaking, it's not a function, although it works like `main()` in C/C++. Maybe describe this as a top-level code block?

I think that language was from the original document I started from. I'm not stuck on it, so any suggestions are welcome, or feel free to just edit it in place. Though if we stick with the principle of "here's how we did it in the SWC workshop", it'd probably be:

```python
def main():
    # things in main

if __name__ == '__main__':
   main()
```

as per [Command-Line Programs - Programming with Python](https://swcarpentry.github.io/python-novice-inflammation/12-cmdline/index.html).

reid-a:
Thanks for linking to the python-novice lesson, I had missed their usage. Even though it's a bit verbose, I think I favor actually making it a main() function and then calling it, following their practice.

This also answered a secondary question I had, about whether just grabbing stuff out of `sys.argv` was a good idea -- it seems hacky, but the obvious alternative requires calling an argument parser, which is a tangential can of worms. I'm pleased to see that python-novice does a good job of explaining sys.argv, so here again, I think echoing the python-novice practice is what we want.

tkphd:
Creating and calling a `main()` function seems un-pythonic. I would prefer keeping the `__name__` guard, and perhaps comment on what that means -- depends on whether the file is invoked directly, or imported.

tkphd:
The scripts in [16-parallel.md](https://github.com/carpentries-incubator/hpc-intro/blob/5888fcc4a9f1724650102ea3c72f6baa575755ea/_episodes/16-parallel.md) call `sys.argv[1]` without first checking the validity of arguments. Recommend including something like

```python
if __name__ == "__main__":
    if len(sys.argv) != 2 or sys.argv[1] == "--help":
        print("Usage: {0} number_of_random_points".format(sys.argv[0]))
        sys.exit()
```

reid-a:
I honestly don't have a hugely strong opinion, but I feel like a point is getting elided.

Both of the recent comments by @tkphd have merit, but they take us away from the path taken in the python-novice lesson that @mikerenfro linked above. 

That's a mainstream Carpentries lesson, so It seems reasonable to at least consider that it's road-tested, and that its maintainers apparently do not believe that the deviations from best-practice compromise their pedagogical goals.

I personally lean very slightly towards "do what python-novice does", but not so that I'd revert the change if it went the other way.

mikerenfro:
The "Adding a Help Message" exercise on the Command-Line Programs page does cover those concepts, but there's a chance it could get skipped in any given workshop. I'm ok to adapt those instructions to this and reference, since at least we're not breaking completely new ground.

tkphd:
Can we compromise, and call the function we call something that isn't `main()`? It feels like a confusing conflation of C and Python concepts.

reid-a:
Fine by me.

tkphd:
Commenting out or deleting the last 3 lines (guard & `main()`) from the code-block centered on ine 530 seems to clear the build error. I have no idea why.

reid-a:
Which file? Commenting out the call to main() in lesson_check.py prevents the check from running at all, but on my check-out of the branch, this is line 568?

tkphd:
Sorry, yeah, in 16-parallel.md.

reid-a:
The lesson check includes a "code-block checker" which apparently requires that all classes in code blocks be in the KNOWN_CODEBLOCKS list, which consists of:
```
# What kinds of code fragments are allowed?
# Below we allow all 'language-*' code blocks
KNOWN_CODEBLOCKS = {
    'error',
    'output',
    'source',
    'warning',
    'language-python',
}
```
Can we tag it `language-python`?

tkphd:
It already is, but the check fails when the final block is there, as if lesson_check is parsing it as two nested blocks?

tkphd:
Hmm, de-indenting from 4 spaces to 2 also seems to resolve the tests, so, definitely a Markdown indentation thing.

reid-a:
Trying to reverse-engineer the code-block check now. One issue is that there is apparently only one code-block "node" in Episode 16, even though the "language-python" tag occurs many times.

reid-a:
Diagnostic tip: You can "manually" parse things via `ruby bin/markdown_ast.rb < _episodes/16-parallel.md`, which gets you the page in JSON format. Only one code-block is identified as such, and it's triggering the failure.

reid-a:
FIX: Do not put blank lines inside the triple-backticks. In particular, on line 542 of episode 16, put a comment character.

This prevents the subsequent part from being ID'd as a "codeblock". The check will pass.

tkphd:
Alternate fix in 312edd3: don't `def main()`, and you can keep the blank lines inside the `__main__` guard.

reid-a:
IMO, this is a dangerous fix. The parser seems to be mis-parsing back-tick blocks with blank lines, and so this approach will likely lead to hard-to-diagnose issues in the future. Commenting the "blank" lines is ugly, but more robust.

tkphd:
Acknowledged. Until Ruby stuff gets fixed upstream, we can use the *other* fix, dea9565.

tkphd:
Even better: the upstream styles repo has updated [`markdown_ast.rb`](https://github.com/carpentries/styles/blob/gh-pages/bin/markdown_ast.rb), fixing all these issues!

mikerenfro:
> Can we compromise, and name the function we call something that isn't `main()`? It feels like a confusing conflation of C and Python concepts.

- "By convention this function is usually called main, though we can call it whatever we want:" -- Command-Line Arguments section of [Programming with Python: Command-Line Programs](https://swcarpentry.github.io/python-novice-inflammation/12-cmdline/index.html)
- [__main__ — Top-level script environmen](https://docs.python.org/3/library/__main__.html)
- [What does if __name__ == “__main__”: do?](https://stackoverflow.com/a/419185) from 2009

The conflation started so long ago, it's basically a tradition now.

I can add a (last?) commit for the checking for command-line arguments. As I'd said before, it's covered in the previous material, but I now realize we'd want to add it to each version of both the serial and MPI codes.

Reactions?
- Thumbs up for merging without the argument check,
- Thumbs down for adding the argument check before merging.

tkphd:
Thanks for the links -- I suppose that ship sailed :unamused: 

The various codes throw `IndexError: list index out of range` if there's no count argument, so we can eschew the explicit checks in favor of brevity.

reid-a:
I think there is an issue with the kramdown parser.

When I run "make lesson-check-all" on my check-out of this branch, I get an error from the requirement for kramdown.
I suspect this is because the relevant "gem add" line has been added to the `.github/workflows/test-and-build.yml` file (so it works for the CI), but has not been added to whatever the makefile uses to load gems, and so fails locally.

A working local test is high-value, IMO.

tkphd:
It worked without further intervention on my end, but my build system is not necessarily clean. Added kramdown-parser-gfm to the Gemfile, give it a try?

reid-a:
Still busted. Seems to be missing from a file, `.vendor/bundle/ruby/2.7.0/gems/jekyll-3.9.0/lib/site_template/_config.yml`, but this is not part of the repo, it's generated at build-time, I think.

tkphd:
When you run `make serve`, it runs `bundle install` first, right? It should populate...

reid-a:
It does run `bundle install`, and pulls in a ton of stuff, but apparently not the right thing. I am also doing clean builds in fresh checkouts, to side-step cache issues, although it's possible that bad state is being cached in my RVM inrastructure (although the high level of activity at serve-time suggests otherwise.)

tkphd:
Happy to see this fixed here & [upstream](https://github.com/carpentries/styles/pull/548). Great work, Andrew!

tkphd:
Hmm, @mikerenfro I thought there had been a commit that fixed the `scatter.png not found` error. Am I mistaken, or was the fix reverted by accident?

_episodes/16-parallel.md
```diff
@@ -40,45 +405,23 @@ in the process. In the context of our queuing system, however, we do not need to
 this information, the `mpirun` command will obtain it from the queuing system, by
```
ocaisa, line 405: `mpirun` needs to be configurable. The standard is actually `mpiexec` but even then there are systems (like ours) that *require* other MPI launchers (`srun`, `aprun`,...)...`{{ site.remote.mpiexec }}`?

_episodes/16-parallel.md
```diff
@@ -125,7 +575,7 @@ Create a submission file, requesting more than one task on a single node:
 {{ site.sched.comment }} {{ site.sched.flag.queue }} {{ site.sched.queue.testing }}
 {% include {{ site.snippets }}/parallel/four-tasks.snip %}
 module load python3
-mpirun ./pi.py 1431652028
+mpirun python pi.py 100000000
```
ocaisa, line 578: This also needs to be configurable:
```
{{ site.remote.mpiexec }} {{ site.remote.mpiexec_args }} {{ site.remote.python }}  pi.py 100000000
```
but this already starts to look a bit tedious, perhaps the jobscript is better off in a snippet

mikerenfro: I've replaced all the `mpirun` commands with `mpiexec` -- I'm good to make it configurable for `srun` or other commands, whether in a snippet or a variable. Could that be a future issue and PR if the remainder of the episode is acceptable?

_episodes/16-parallel.md
mikerenfro: I've replaced all the `mpirun` commands with `mpiexec` -- I'm good to make it configurable for `srun` or other commands, whether in a snippet or a variable. Could that be a future issue and PR if the remainder of the episode is acceptable?

_episodes/16-parallel.md
mikerenfro: I'm good for either option, as mentioned elsewhere, but if we can get this merged in and decide on that in a separate issue/PR, that'd help.

mikerenfro:
@ocaisa @tkphd @reid-a et al., are we good to merge these in as a starting point?

_episodes/16-parallel.md
```diff
@@ -125,7 +575,7 @@ Create a submission file, requesting more than one task on a single node:
 {{ site.sched.comment }} {{ site.sched.flag.queue }} {{ site.sched.queue.testing }}
 {% include {{ site.snippets }}/parallel/four-tasks.snip %}
 module load python3
-mpirun ./pi.py 1431652028
+mpiexec python pi.py 100000000
```
ocaisa, line 578: I'm sorry to be picky about this but for this to run on four tasks you are relying on the MPI runtime being smart and picking up information for the resource scheduler. To be clear we should also give the necessary argument to tell it how many tasks to use:
```suggestion
mpiexec -n 4 python pi.py 100000000
```

tkphd: OK,  we've passed deep into "make this a snippet" territory.

mikerenfro: Understood, and that's mentioned back in the body text about the scheduler handling it how many tasks to start. Might not be a universal rule, but it's more brittle to have the user specify the number of tasks to run, and have to keep that in sync with the `--ntasks` or similar flags in their resource requests.

_episodes/16-parallel.md
```diff
@@ -90,21 +433,149 @@ clearly. Press "q" to exit the pager program (`less`).
...
+   my_pi = 4.0 * sum(counts) / sum(partitions)
+   end_time = datetime.datetime.now()
+   elapsed_time = (end_time - start_time).total_seconds()
+   size_of_float = np.dtype(np.float64).itemsize
+   memory_required = 3 * sum(partitions) * size_of_float / (1024**3)
+   print("Pi: {}, memory: {} GiB, time: {} s".format(my_pi, memory_required, elapsed_time))
+```
+{: .language-python}
+
+The resulting MPI Python script is:
```
tkphd, line 532: Once more, let's not print the whole thing. Perhaps links for the instructor to the completed script, in case someone has fallen hopelessly behind, but we want live-coding rather than copy-paste.

_episodes/16-parallel.md
```diff
@@ -142,4 +613,66 @@ output file, and examine it. Is it what you expected?
 
 * How good is the value for &#960;?
 * How much memory did it need?
-* How much of that memory was used on each node?
+* How long did the job take to run?
```
tkphd, line 616: * How much faster was this run than the largest serial job we ran?

_episodes/16-parallel.md
```diff
@@ -142,4 +613,66 @@ output file, and examine it. Is it what you expected?
...
+the workload, the total run time cannot be less than one hour.
+
+In practice, it's common to evaluate the parallelism of an MPI program by
+* running the program across a range of CPU counts,
+* recording the execution time on each run,
+* comparing each execution time to the time when using a single CPU.
+
+The speedup factor *S* is calcuated as the single-CPU execution time divided
+by the multi-CPU execution time.
+For a laptop with 8 cores, the graph of speedup factor versus number of cores
```
tkphd, line 655: The specifics of this can get tripped up in *cores* vs. *threads*: there'll be a severe drop at the `cores+1` point, rising back up to `cores` performance at `threads`.

_episodes/16-parallel.md
```diff
@@ -142,4 +613,66 @@ output file, and examine it. Is it what you expected?
...
+used shows relatively consistent improvement when using 2, 4, or 8 cores, but
+using additional cores shows a diminishing return.
+
+{% include figure.html url="" max-width="20%" file="/fig/laptop-mpi_Speedup_factor.png" alt="MPI speedup factors on an 8-core laptop" caption="" %}
```
tkphd, line 659: 20% is not legible.

tkphd (changes requested):
Valuable contribution @mikerenfro, thank you for all the work that went into it!
A few things I would prefer to see rephrased.
Overall, 20% figure widths are much too small. Try 30 to 40%, but use your judgment and a local build to tune.

_episodes/16-parallel.md
ocaisa: I do think this is best served by a snippet, that leaves people the scope to comment the code as necessary, as well as more straightforward to tweak. 

It's definitely brittle to have to specify the tasks given it has to be done in context, but there are many use case for MPI outside of a scheduler and it would be required there. In any case, I think it's a case of allowing for whatever the instructor is comfortable with, so a snippet will provide that flexibility without us having to make these calls.

_episodes/16-parallel.md
mikerenfro: Maybe I adjust the graphs to read "number of ranks" on the horizontal axis instead of "number of cores" or threads. If ranks<=cores, should get consistent speedup. If ranks>cores, diminishing returns. If ranks>threads, should be no improvement at all.

_episodes/16-parallel.md
```diff
@@ -16,6 +16,371 @@ We now have the tools we need to run a multi-processor job. This is a very impor
...
+    counts = inside_circle(n_samples)
+    my_pi = 4.0 * counts / n_samples
+    print(my_pi)
+
+if __name__ == '__main__':
+    main()
+```
+{: .language-python}
+
+The entire Python script is:
```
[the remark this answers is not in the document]
mikerenfro: Currently, the instructors' note page just has a glossary. If you want to modify it so there's a place for other notes, I can add this.

_episodes/16-parallel.md
```diff
@@ -32,56 +397,34 @@ this is a common tool on HPC systems.
 {: .callout}
 
 MPI jobs cannot generally be run as stand-alone executables. Instead, they should be
```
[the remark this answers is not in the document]
ocaisa: We actually ran into this a few times and in fact there's nothing in the standard that requires that a a parallel program can run outside of an MPI environment. We've had this issue with both Intel MPI and ParaStation MPI at one point, usually they still fix this though because it is clearly useful.

tkphd (approved):
Solid contribution, thank you @mikerenfro!

PR #309, opened 2021-02-19
hostname should give a consistent answer

ocaisa:
see https://github.com/carpentries-incubator/hpc-intro/blame/gh-pages/_episodes/13-scheduler.md#L125


tkphd (approved):
Good catch!

PR #311, opened 2021-02-19
No line ending makes job cancellation message hard to read

ocaisa:


ocaisa:
Sorry, I copy pasted the wrong file contents

ocaisa:
Sorry, got confused, this PR is still valid

_episodes/13-scheduler.md
```diff
@@ -314,8 +314,8 @@ minutes.
 {{ site.sched.comment }} {{ site.sched.flag.time }} 00:00:30
 
 echo -n "This script is running on ... "
-sleep 120 # time in seconds
 hostname
+sleep 120 # time in seconds
```
ocaisa, line 318: Either this reordering or remove the `-n` option to `echo`, leaving it resulting in mega long line

tkphd:
The disorderly output is deliberate. We want it to be clear that the job began but, due to hitting the time limit, failed to complete. Having the script print before sleeping, the only indication that it didn't finish would be in the queuing system, which is not as intuitive.

Alternative ideas are welcome.

ocaisa:
Ok, but then I would drop the `-n` on the echo, the resulting snippet is just one big line. Will I do that?

ocaisa:
I'm ok, if this is closed, but for me the current setup makes it hard to actually read the cancellation message

tkphd (approved):
LGTM. Thanks for proofreading the outputs!

PR #313, opened 2021-02-22
Some minor updates

ocaisa:
* Serving the content locally creates a `.bundle` directory in the root that should not be tracked, added this to `.gitignore`
* We use `{: .warning}` somewhere but this is not valid lesson syntax and does not render, replace with `{: .callout}` and an appropriate heading
* For lesson rendering in GitHub, instructions are to copy-paste from site configuration into `_config.yml`. Episode ordering is part of site configuration so should appear alongside everything else

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -13,3 +13,4 @@ files/*.tsv
 Gemfile.lock
 scratch/
 .vendor
+.bundle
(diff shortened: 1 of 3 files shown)


ocaisa:
CI is failing due to problem downloading the remote theme that just started occurring

psteinb (commented):
Thanks. LGTM. We should merge this once CI passes.

tkphd (approved):
Yep, good fixes!

PR #314, opened 2021-02-22
Fix broken link to learner profiles

ocaisa:
In https://github.com/carpentries-incubator/hpc-intro/pull/269 we removed the "Learner Profiles" file, with the reasoning that it was covered in the introduction of the first episode.

Personally, I liked the fact that that file was learner-focussed as opposed to task-focussed (which is how I would read the first few paragraphs of the introductory episode). My personal preference would be to actually reintroduce the file and leave the (currently broken) link in the main README.md unchanged. I'm open to discussing it though.

### DIFF
--- a/README.md
+++ b/README.md
@@ -121,7 +121,7 @@ building your actual lesson, to re-order these files, or omit one or more of the
 
 * Prelude
     * [User profiles](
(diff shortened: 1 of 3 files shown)


tkphd:
Sure, I'm on board!

tkphd (approved):
LGTM!

PR #316, opened 2021-02-23
Use language-bash not bash in formatting

### LINKED ISSUE #315: .bash instead of .language-bash
I was looking at https://github.com/hpc-carpentry/hpc-shell/pull/35 and I noticed that we are using `{: .bash}` right throughout this lesson, but this is not actually part of the [Carpentries syntax](https://carpentries.github.io/lesson-example/04-formatting/index.html#formatting-code) and is actually rendering like `{: .source}`. We should be using `{: .language-bash}`

ocaisa:
Fixes #315 

`{: .bash}` is not actually part of the [Carpentries syntax](https://carpentries.github.io/lesson-example/04-formatting/index.html#formatting-code) and is actually rendering like `{: .source}`. We should be using `{: .language-bash}`.

### DIFF
--- a/_episodes/12-cluster.md
+++ b/_episodes/12-cluster.md
@@ -82,7 +82,7 @@ username and the remote computer you can reach from the outside world, {{ site.r
 ```
 {{ site.local.prompt }} ssh {{ site.remote.user }}@{{ site.remote.login }}
 ```
-{: .bash}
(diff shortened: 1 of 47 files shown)


psteinb (approved):
great PR. thanks.

mikerenfro (approved):

PR #322, opened 2021-02-25
Also check snip files for syntax

ocaisa:


ocaisa:
I can't get the CI to pass on this, I don't see anything wrong with the snippet file. I'm leaning towards adding an exclusion to the checker!

reid-a:
@ocaisa  Maybe obvious, but it feels like the key has to be in how your snippet differs from the others?

I think the failing check doesn't even use the snippets, it uses the defaults.

Are you able to run "make lesson-check-all" by hand in your check-out to trace the error?

ocaisa:
It's not my snippet file, it's the one at https://github.com/carpentries-incubator/hpc-intro/blob/gh-pages/_includes/snippets_library/UCL_Myriad_sge/modules/wrong-gcc-version.snip

The tests pass if I remove the first 32 lines, it's just bizarre.

ocaisa:
The failing test is the one that includes running `make lesson-check`, which does check all the snippet files

ocaisa:
Ok, I found it it is lines like https://github.com/carpentries-incubator/hpc-intro/blob/gh-pages/_includes/snippets_library/UCL_Myriad_sge/modules/wrong-gcc-version.snip#L10

reid-a:
Looks promising!
FYI it's possible to run the Markdown parser "manually" and inspect the resulting JSON output -- that's how I resolves a similar issue recently. The command is issued from line 188 of `bin/util.py`.

ocaisa:
I don't get it you have exactly the same type of lines in https://github.com/carpentries-incubator/hpc-intro/blob/gh-pages/_includes/snippets_library/ComputeCanada_Graham_slurm/modules/available-modules.snip#L2

I'll see if the json is more helpful

ocaisa:
I run
```
bundle exec ruby bin/markdown_ast.rb < _includes/snippets_library/UCL_Myriad_sge/modules/wrong-gcc-version.snip | less
```
but I don't really know what I'm looking for, according to that json  it looks fine to me. If I remove the lines, the only difference in the json files is the missing lines...except that with the lines removed the tests pass :(

tkphd:
I think this is to do with `lesson_check.py:572`, where it sets `CheckNonJekyll` for the `.snip` files. Setting `CheckGeneric`, then inserting metadata into the problem snippet, clears that lone file. All the rest throw missing metadata errors.

Inserted metadata:

    ---
    layout: page
    ---

This does not seem to affect the rendered output when the site is built.

ocaisa:
It's totally weird, the only thing that `CheckNonJekyll` does not do is not check for metadata. Otherwise they are identical (`CheckNonJekyll` is not a great name).

_includes/snippets_library/UCL_Myriad_sge/modules/wrong-gcc-version.snip
```diff
@@ -7,7 +7,7 @@ e.g.
 {: .language-bash}
  
 ```
-------------------------------------- /shared/ucl/apps/modulefiles/applications --------------------------------------
+~~~~~~~~~~~~~~~~~~~~~~ /shared/ucl/apps/modulefiles/applications ~~~~~~~~~~~~~~~~~~~~~~
```
ocaisa, line 10: This is a complete hack but I'm not willing to spend more time trying to debug `lesson_check.py`

bin/lesson_check.py
```diff
@@ -558,7 +570,8 @@ def __init__(self, args, filename, metadata, metadata_len, text, lines, doc):
     (re.compile(r'index\.md'), CheckIndex),
     (re.compile(r'reference\.md'), CheckReference),
     (re.compile(r'_episodes/.*\.md'), CheckEpisode),
-    (re.compile(r'.*\.md'), CheckGeneric)
+    (re.compile(r'.*\.md'), CheckGeneric),
+    (re.compile(r'.*\.snip'), CheckNonJekyll)
```
ocaisa, line 574: I deliberately chose `CheckNonJekyll` otherwise we would have to add yaml metadata to all the snippets (which make no sense)

reid-a:
For what it's worth, I believe the checker's upstream is still actively being developed, so the situation may improve on its own at some point. If we could nail down the misbehavior here, that would help, but IMO that's a ton of work with the wrong focus for us.

tkphd (approved):
This is a solid contribution. Thank you for all your troubleshooting effort, @ocaisa! 
These changes ensure that codeblocks in snippets are properly tagged. I'm comfortable with the "hack" of using a different character for the frames on module output. `lesson_check.py` is kind-of nightmarish, and we're diverging from upstream yet again, but that's fine, we're grown ups.

PR #323, opened 2021-02-25
Wrap files at 79 characters for cleaner diffs and readability

### LINKED ISSUE #282: wrap markdown at 79 columns
The GitHub `diff` interface wraps lines at 80 characters. Since we use 100 characters as the max width, this makes for ugly `diff`s with lots of continued lines. To improve the clarity of these diffs, as well as readability in terminal text editors, we should back away from this edge, and wrap our content at 79 characters.

This is easily done with emacs, but the change would touch almost every file in the repository, so this should be coordinated with other efforts to minimize merge conflicts and the number of branch rebases to be done.

tkphd:
Edits to all the Markdown. Fixes #282.


CONTRIBUTING.md
```diff
@@ -1,107 +1,121 @@
...
+workshop, so we are usually *not* looking for more concepts or tools to add to
+them. As a rule, if you want to introduce a new idea, you must (a) estimate how
+long it will take to teach and (b) explain what you would take out to make room
+for it. The first encourages contributors to be honest about requirements; the
+second, to think hard about priorities.
 
-We are also not looking for exercises or other material that only run on one platform. Our workshops
-typically contain a mixture of Windows, macOS, and Linux users; in order to be usable, our lessons
-must run equally well on all three.
+We are also not looking for exercises or other material that only run on one
```
mikerenfro, line 81: Not sure this paragraph is as applicable to HPC Carpentry as it would be to SC, DC, or LC. Most of what we talk about is on a Linux HPC, with the exception of the Pi-calculating Python code that starts out on the local computer.

reference.md
```diff
@@ -9,35 +9,46 @@ title: Knowledge Base
...
+meaning. When people say "Kilobyte", they mean 1024 Bytes instead. In that
+spirit, a Megabyte is 1024 Kilobytes.
 
 To address this ambiguity, the [International System of Quantities](
-https://en.wikipedia.org/wiki/International_System_of_Quantities) standardizes the *binary* prefixes
-(with base of 2<sup>10</sup>=1024) by the prefixes Kibi (ki), Mibi (Mi), Gibi (Gi), etc. For more
-details, see [here](https://en.wikipedia.org/wiki/Binary_prefix).
+https://en.wikipedia.org/wiki/International_System_of_Quantities) standardizes
+the *binary* prefixes (with base of 2<sup>10</sup>=1024) by the prefixes Kibi
+(ki), Mibi (Mi), Gibi (Gi), etc. For more details, see
```
mikerenfro, line 34: Mebi, as per https://en.wikipedia.org/wiki/Binary_prefix#IEC_prefixes

reference.md
```diff
@@ -9,35 +9,46 @@ title: Knowledge Base
...
 ### "No such file or directory" or "symbol 0096" errors
 
-`scp` and `rsync` may throw a perplexing error about files that very much do exist. One source of
-these errors is copy-and-paste of command line arguments from Web browsers, where the double-dash
-string `--` is rendered as an em-dash character `—` (or en-dash `–`, or horizontal bar `―`). For
-example, instead of showing the transfer rate in real time, the following command fails mysteriously.
+`scp` and `rsync` may throw a perplexing error about files that very much do
+exist. One source of these errors is copy-and-paste of command line arguments
+from Web browsers, where the double-dash string `--` is rendered as an em-dash
+character `—` (or en-dash `–`, or horizontal bar `―`). For example, instead of
```
mikerenfro, line 42: Do we want the HTML-ized versions of these characters that we typically use elsewhere, or the unicode versions that we have here?

CONTRIBUTING.md
tkphd: For now, CONTRIBUTING is boilerplate taken from The Carpentries. It would be worth editing for a better fit, but that discussion belongs in a separate issue.

reference.md
tkphd: 1f9b42d

reference.md
tkphd: cdbf42e

mikerenfro (approved):

tkphd:
142 files changed... did I miss any!!?

mikerenfro (approved):
Via [Unix StackExchange: Find any lines exceeding a certain length](https://unix.stackexchange.com/a/42899/8967) and

```
for f in $(find . -type f -name '*.md' -o -name '*.snip'); do echo $f; perl -nle 'print if length$_>79' $f; done | less
```

I don't see anything in the `*.md` or `*.snip` files over 79 characters that is an issue (just links and things that can't be fixed due to length of variable names that all have to be on one line).

PR #157, opened 2020-07-17
Adds a shell flag to the properties

dpshelio:
This includes #156 as I couldn't demonstrate the need without it. We can rebase it once #156 is merged.

Essentially, at UCL we need `#!/bin/bash -l` to use our profiles (this was already discussed in #126 but  then the lessons were restructured). The way this includes this option (or other in the future) is by adding as an extra variable.

PR #324, opened 2021-02-26
Make bash shebangs configurable

ocaisa:
Also replaces #157


tkphd (approved):
Looks good to me! This change also makes for easy switching, for example where `#!/bin/bash` is preferable. Thanks @ocaisa!

PR #327, opened 2021-03-04
center & shrink figure captions

tkphd:
The `{% include figure.html ...` method has a "caption" field, but text entered there is rendered inline with the surrounding text, creating confusion. This PR centers and shrinks the caption, to make it more visually recognizable as such.


ocaisa:
Tested this, works fine, thanks @tkphd !

ocaisa (approved):

PR #331, opened 2021-03-05
use `make` instead of reproducing `codespell` command

tkphd:


reid-a (approved):
Looks good. Also enjoying the thematic consistency of the spelling fixes.

Issue #306, opened 2021-02-15
2021-02 review of lesson text, 18-responsibility


mikerenfro:
Request for comment:

## Be kind to the login nodes

1. Are interactive queues common in HPC environments? I know I have them, as does PSC. Would be more site-specific stuff for the snippets library, but would also give a better option than testing on the login node.
2. `&emdash;` doesn't get rendered as HTML, the ampersand is converted to `&amp;`.
   
## Transferring data

1. Adjust caption to match colors used in figure. Grey and blue used in the figure, but black and white in the caption.

reid-a:
The current wording about use of the login nodes avoids possibly-extraneous detail about various systems and their configurations, and so revolves around "be respectful" and "ask your site administrator" as primary anchors. We could mention the possibility of interactive queues, as something your sysadmin might know about, but I'd stop short of recommending them, because we can't promise they'll be available.

I am not seeing the emdash and amp issues, can you locate them more precisely?

Agreed about the figure caption.

reid-a:
PR #334 is a fix to the caption.

PR #334, opened 2021-04-01
Fix to the network diagram caption.

reid-a:


reid-a:
Partial fix for issue #306.

mikerenfro (approved):

PR #337, opened 2021-04-05
Glossary improvements

harivyasi:
Two commits. Done as part of instuctor checkout process.


reid-a (approved):
This is a welcome addition, and hopefully will get the ball rolling on additional entries. Thanks!

PR #341, opened 2021-04-18
Create parallel-fortran example

bkmgit:
It would be helpful to have examples in a number of programming languages, to allow choice and comparison.
This is for Fortran, but other languages would be nice to add as well.


tkphd:
As a practitioner, I find this very cool, and welcome the contribution.
For our learners, the focus of *this* lesson should be on using the HPC machinery, and the high-level Python language is the best fit for that constraint. I suggest we begin planning an "HPC Programming" lesson and consider moving this Fortran code there.

bkmgit:
- We should not force everyone to use Python, while it is useful for scripting and increasing programmer productivity, it has serious drawbacks for efficiency, see for example https://ioinformatics.org/journal/INFOL026.pdf 
- Allowing people to contribute different parallel versions is the way to go
- The artificial memory utilization in the Python program is needed in Python because loops without automatic numpy vectorization are slow. Tradeoffs of speed for memory are common in many algorithm implementations, but probably not needed in all the implementations.

tkphd:
I completely agree, @bkmgit: I use C++ & CUDA for my HPC work, with some Python glue. We are not trying to force people to use Python, and certainly are not suggesting people use vanilla Python for high-performance applications.

Fundamentally, the HPC Intro lesson is not about the code, it's about the infrastructure: how to connect, plan & launch jobs, and manage file inputs & outputs. Python, as a high-level and reasonably human-readable language, allows us to focus on the pedagogical goals without getting bogged explaining the language. Fortran achieves its performance because of the fact that it is a low-level language. I could explain the code you submitted to newbies, but I really wouldn't want to unless that was the only teaching goal. Hence my suggestion to partition it into a separate, code-centric, lesson. It's possible we could provide the Fortran version as an alternative episode, so if you know your learners are more familiar with Fortran than they are with Python, you could teach to their strengths. Given the rise of Python over the past decade, I'm not sure how realistic the notion is, but it's not impossible.

There has been a lot of discussion around the curriculum and teaching goals of HPC Intro. We would do well to revisit the over-arching curriculum of our lessons as a whole, and figure out how to smoothly chain from one lesson to the next in building a workshop. Your experience teaching the material is high value, and I look forward to ironing out the designs with you.

bkmgit:
Expect to test out:

- C
- C++
- Java
- Scala
- Go
- R
- Chapel
- Julia
- Rust
- Perl
- Lua 

One can even use map-reduce for this problem, which may be helpful for people using big data frameworks

bkmgit:
The aim should be that course participants can choose the language they are most familiar with for execution, with instructor using a common high level explanation.

tkphd:
The conventional Carpentries approach is to have the Learners and Instructor typing the same thing, to remain engaged and get used to keystrokes. For an individual following a web tutorial, it could make sense to offer all these options for comfort, but the Carpentries exists because that style of learning is not as effective as a guided tutorial working through common material. 

I see that you're passionate about providing options and helping people learn in their preferred languages. Have you received strong feedback from learners that Python is unacceptable?

bkmgit:
So we can do an effective test of this, it would be great to add templates for code blocks of
- language-fortran
- language-c
- language-cpp
- language-java
- language-scala
- language-go
- language-r
- language-chapel
- language-julia
- language-rust
- language-perl
- language-lua

Where can these be added?

tkphd:
`KNOWN_CODEBLOCKS` in `bin/lesson_check.py`

tkphd:
Recommend making this WIP Ep. 19, so the "consecutive numbering" check can pass.

bkmgit:
Is it possible to have 16 which has Python as default and gives links to other options, or just lists all the possible language options, and then 16-a 16-b etc for other language possibilities?

tkphd:
No. Lessons must be numbered consecutively.

tkphd:
If you want options, the episode would have to be snippetized.

reid-a:
> No. Lessons must be numbered consecutively.

Possibly a minor sidebar, but I think lessons can be nonconsecutive since the merger of PR #263 , closing issue #230. Actual nonsequential lessons were never implemented, but it's now a policy question, and not a technical question.

reid-a:
WRT this issue more broadly, I fear that the introduction of parallel examples in additional languages, while interesting and valuable, detracts from the focus of HPC Intro as a whole, which is primarily about how to execute code in a cluster environment with shared resources. 

The primary reason for including parallel code at all (which, if I recall, was not uncontested) is that access to parallel resources is one of the principal features of cluster HPC environments, and is a big part of the motivation for learners in undertaking this lesson in the first place. The absence of a parallel example undercuts this motivation. 

The choice of a Python example (over the previous C "cpi" example) was motivated primarily by the relative simplicity of the code, and the relatively high level of the abstractions used, with the hope that it would not be too much of a distraction from the focal point of the lesson.

Parallel programming and parallel models (especially MPI and OpenMP) are obviously valuable to the community, but I remain concerned that taking learners down that particular rabbit-hole in this lesson leave them confused about *both* operational issues with clusters and parallel programming. 

I would enthusiastically endorse a more separable lesson on parallel programming specifically, possibly as a successor to both "parallel novice" and "hpc intro".

bkmgit:
Ok, the snippets option may be good here. Most procedural languages would have a very similar format.

tkphd:
The lesson has diverged from the pedagogical approach of this PR.

PR #342, opened 2021-04-18
Update lesson_check.py

bkmgit:
Enable use of a wider range of programming languages

Try to fix build error:

https://github.com/carpentries-incubator/hpc-intro/pull/341/checks?check_run_id=2373944669#


bin/lesson_check.py
```diff
@@ -91,6 +91,17 @@
     'language-r',
     'language-shell',
     'language-sql',
+    'language-fortran',
```
tkphd, line 94: Good additions. Please alphabetize the list.

tkphd (changes requested):

bin/lesson_check.py
bkmgit: done, thanks for the suggestion

bin/lesson_check.py
```diff
@@ -80,30 +80,30 @@
...
-    'language-scala',
-    'language-go',
-    'language-chapel',
-    'language-julia',
-    'language-rust',
-    'language-perl',
-    'language-lua',
-    'bash',
+    'output',
     'python'
```
tkphd, line 105: Need a comma after 'python'

tkphd (changes requested):

tkphd (approved):
Looks good to me! Thanks for the patches.

bin/lesson_check.py
bkmgit: Thanks.

ocaisa:
Does this mean the theme can actually syntax highlight these languages? Did you test all these?

tkphd:
Based on [`lesson.scss`](https://github.com/carpentries/styles/blob/gh-pages/assets/css/lesson.scss), these should render in colored boxes, but will not have the language name printed above until/unless they are all added.

ocaisa:
Ok, but if they are all second class citizens in this context how do we make clear to anyone who wants to use them?

tkphd:
* Is there a way to create an add-on SCSS file to provide complete coverage for our repos?
* The SCSS matches `language-*` to apply the colorbox, but explicitly writes labels for a handful of "supported" languages. Could the pattern matching be updated to both build the colorbox *and* use the `*` as the box title?
* The syntax highlighters, Rouge and Pygment, support all these languages and more, so the restriction is arbitrary.

tkphd:
(implicitly, instead of warding off developers from using these syntax flags, I'm suggesting updating the upstream machinery so it "just works")

bkmgit:
Can add these. Thanks for the suggestion.

tkphd:
The Fortran code renders like this. "Code" is not untrue, and it looks like the syntax highlighter sorted out what to do. I don't think this requires a disclaimer, though the upstream improvements would still be welcome.

tkphd:
@bkmgit, we lean on [carpentries/styles](https://github.com/carpentries/styles) for our CSS, since we really don't want to maintain the back-end at that level unless it's really necessary. Please don't add CSS to this repo, unless the above rendered code block is unacceptable.

bkmgit:
Thanks for the suggestion -  https://github.com/carpentries/styles/pull/580

tkphd:
:tada: @bkmgit's [PR](https://github.com/carpentries/styles/pull/580) was merged!

PR #344, opened 2021-04-19
Add discussion on random number generation

bkmgit:


reid-a (approved):
This is a useful addition, and a potentially important point. Possibly better wording than "regular functions" might be found later on, since it reads to me like a mathematical sub-set of fucntions, but I think in this context actually means "default system or language libraries".

This is not an obstacle to acceptance.

tkphd:
I'm concerned that, while a good thing to point out in general, the caveats of pseudo-random number generators is diving into too much detail, and is not within the scope of this lesson.

1. The default is acceptable.

    From the [NumPy documentation](https://numpy.org/neps/nep-0019-rng-policy.html), `numpy.random` uses the venerable [Mersenne Twister](https://en.wikipedia.org/wiki/Mersenne_Twister). For this lesson, that's "good enough" randomness.

2. A parallel PRNG is not necessary.

   In general, "parallel" RNG can be modeled as an ensemble of serial RNGs. It only gets tricky when
   * the parallel tasks are not independent (e.g., shared threads clobbering state variables at global scope), 
   * when every parallel task must generate the *same sequence*, or
   * when a constant seed is hard-coded, without modification per-rank.

   For this example, Monte Carlo with pure MPI and no explicit PRNG seed, none of these conditions applies.

Is there a pedagogical reason to discuss the complexities of PRNG here?

reid-a:
Random number generation is notoriously difficult to do well, so a cautionary note about that is not amiss, in my opinion. 

I am having some difficulty digging up citations, but I recall that there is a Knuth essay on the surprisingly small useful-parameter subspace of the linear congruential method, and that there were grave concerns about the quality of the "Numerical Recipes" PRNGs when that book was published by SIAM luminaries in 1986 -- the observed variance of the numerical sequence was far from the statistically-expected variance, which had implications for e.g. Monte Carlo simulations which might be tempted to use it.

reid-a:
Citatoins: [Knuth](https://www.informit.com/articles/article.aspx?p=2221790), especially, "random numbers should not be generated with a method chosen at random", and [Numerical Recipes skepticisim](https://www.uwyo.edu/buerkle/misc/wnotnr.html).

tkphd:
All of us who have written code in lower-level languages have been bit by shoddy PRNGs. In C, `rand()` remains a terrible choice. In a course on any such programming language, significant time should be dedicated to making wise decisions, PRNGs being one of those.

This is not a programming course. We chose Python because it abstracts away the low-level fidgety details and, for the most part, implements reasonable defaults somewhere out of sight. NumPy makes very good decisions for us, and if we don't mess with it, we're fine.

If we want to jump into this, let's teach a good practice (e.g., setting the random seed so we get reproducible results) that breaks in parallel (same seed on every rank means the estimate is identical to the serial case). That opens a logical point to discuss why using random numbers in parallel can be tricky, but it comes at a later point where the discussion will not distract attention from learning the serial algorithm to begin with. The simple fix is adding (or modulo dividing) the MPI rank to the constant seed. Even when the equivalent of `time(NULL)` is used, adding the rank is important to get different numbers.

I'm certainly not opposed to mentioning the quirks of PRNGs, but there has to be a good reason, and placement of the discussion should make sense, not confusion.

ocaisa:
> If we want to jump into this, let's teach a good practice (e.g., setting the random seed so we get reproducible results) that breaks in parallel (same seed on every rank means the estimate is identical to the serial case). That opens a logical point to discuss why using random numbers in parallel can be tricky, but it comes at a later point where the discussion will not distract attention from learning the serial algorithm to begin with. The simple fix is adding (or modulo dividing) the MPI rank to the constant seed. Even when the equivalent of `time(NULL)` is used, adding the rank is important to get different numbers.
> 
I really like this idea. The explicit seed is used already in the hpc-parallel-novice equivalent of this example and triggered this exact discussion at the last workshop.

mikerenfro (commented):
No arguments with any of the above comments or recommendations.

_episodes/16-parallel.md
```diff
@@ -73,6 +73,16 @@ def inside_circle(total_count):
 ```
 {: .language-python}
 
+> ## Random Number Generation
+> 
+> For simplicity, random numbers are generated using regular functions, if
```
reid-a, line 78: I think the wording here needs to be clearer -- "random numbers are generated" refers to the preceding example code, not the general case.

I propose: "In the preceding code, random numbers are conveniently generated using the built-in capabilities of NumPy. In general, random-number generation is difficult to do well, it's easy to accidentally introduce correlations into the generated sequence. If high quality random numbers are needed..."

reid-a (changes requested):

_episodes/16-parallel.md
bkmgit: Thanks. Made the modification.

bkmgit:
@tkphd @mikerenfro Thanks for your suggestions. Have removed references to the parallel number generation libraries since they do not have Python wrappers. At the moment, the discussion is at the end of the serial implementation. Adding a parallel implementation based on what is described at:
- https://albertcthomas.github.io/good-practices-random-number-generators/
- https://towardsdatascience.com/stop-using-numpy-random-seed-581a9972805f
- https://numpy.org/neps/nep-0019-rng-policy.html

reid-a:
It looks like the discussion about random numbers is now duplicated, occurring twice in the file? I haven't built this version, so maybe it still works in the built version, but it looks weird, is there a reason for this?

bkmgit:
@reid-a Sorry, haste in committing. Removed duplicate.

reid-a (approved):

reid-a:
Looks good, thanks! 

I have re-requested review from @tkphd, which appears to be necessary to proceed. I have approved this wrt the changes I requested.

_episodes/16-parallel.md
```diff
@@ -111,6 +113,18 @@ If we run the Python script locally with a command-line parameter, as in
 ```
 {: .language-bash}
 
+> ## Random Number Generation
```
tkphd, line 104: Erroneous duplication of the RNG discussion box

_episodes/16-parallel.md
```diff
@@ -74,13 +74,15 @@ def inside_circle(total_count):
...
 
 > ## Random Number Generation
-> 
-> For simplicity, random numbers are generated using regular functions, if
-> a high quality random numbers are needed in parallel applications, one should
-> use libraries such as [SPRNG](http://sprng.org/) and 
-> [Kmath-random](https://www.r-ccs.riken.jp/labs/lpnctrt/projects/kmath-random/).
-> * Discuss why generating high quality random numbers in parallel might be 
-> difficult.
+> In the preceding code, random numbers are conveniently generated using the
```
tkphd: Better phrasing than we had before! For the benefit of Instructors, we should include a Solution sub-box with answers to the questions posed here. 

```markdown
> > ## Solution
> >
> > * Computers are deterministic
> > * Yes
> > * This is not a suitable discussion question, as it is addressed to a self-taught learner rather than the classroom (virtual or physical).
> > {: .solution}
```

tkphd (changes requested):
Looking a lot better.

_episodes/16-parallel.md
bkmgit: This has been removed.

_episodes/16-parallel.md
bkmgit: Thanks. Suggestions implemented.

tkphd (approved):
Looks great, thanks @bkmgit !

PR #346, opened 2021-04-25
Use curl instead of wget

bkmgit:
Curl is installed on GitBash but wget is not by default, so it is helpful to use curl as it will make the lesson easier to use from a variety of shell environments.

### DIFF
--- a/_episodes/15-transferring-files.md
+++ b/_episodes/15-transferring-files.md
@@ -7,7 +7,7 @@ questions:
 objectives:
 - "Be able to transfer files to and from a computing cluster."
 keypoints:
-- "`wget` downloads a file from the internet."
+- "`wget` and `curl -O` download a file from the internet."
 - "`scp` transfer files to and from your computer."
 - "You can use an SFTP client like FileZilla to transfer files through a GUI."
 ---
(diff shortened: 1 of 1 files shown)


tkphd:
`wget` is installed by default in Linux environments, but `curl` is not. We'll need a better solution than only using one or the other.

_episodes/15-transferring-files.md
```diff
@@ -19,16 +19,17 @@ cover here.
 
 ## Download files from the Internet
 
-One of the most straightforward ways to download files is to use `wget`. Any
+One of the most straightforward ways to download files is to use `curl`.  Any
```
mikerenfro, line 22: Remove extra spaces around `Any`?

mikerenfro (changes requested):
No arguments in principle, since both Git Bash and RHEL-derived OSes install `curl` by default, but not `wget`. Not sure how prevalent other non-RHEL OSes are in HPC, or which ones include `wget` but not `curl` by default.

Maybe there's room for a callout or other note where we tell the learner that the equivalent to `curl -O` is `wget`? They could reasonably Google it or use man pages, but might be worth adding it explicitly.

None of the above would block merging, so I'm only marking this as "Request changes" for the one set of extra spaces on line 22. If I can edit that myself, I do so, then mark this as "Approved", and wait a day or so for others' review.

mikerenfro (approved):

mikerenfro:
> 
> 
> `wget` is installed by default in Linux environments, but `curl` is not. We'll need a better solution than only using one or the other.

Did not see this before I started my review. Will verify this one way or the other, since that's not my recalled experience.

tkphd:
Amended my comment: CTCMS uses Debian extensively, where `curl` is not installed by default. 
I'm not opposed to switching up the lesson, but we would have to make a note somewhere -- perhaps in a Workshop Preparations checklist -- to ensure `curl` is installed beforehand, or create a callout for "`curl` not working? Try `wget`!"

tkphd:
Hmm, since these commands are running on the local device (laptop, etc), we have no control over package installation. 
We should add a note in Setup for users to check `which curl` and install if absent.

mikerenfro:
> Hmm, since these commands are running on the local device (laptop, etc), we have no control over package installation.
> We should add a note in Setup for users to check `which curl` and install if absent.

It's mostly on the local laptop, but there's one instance on the HPC, where we're checking if they can download files directly, or if the system has been firewalled off from direct downloads.

tkphd:
Right. We need workshop (#347) and local(#348) configuration guidance.

bkmgit:
Modifying to include both curl and wget

tkphd:
Thank you, @bkmgit -- I appreciate all the work you're putting into the lesson!

PR #355, opened 2021-05-07
Fixing extra spaces after sbatch commands

mikerenfro:
Extra spaces due to the `site.sched.submit.options` variable being empty for many of the sites. This does clutter up the templates a bit, but not sure if we have any better options other than living with the extra spaces.

### DIFF
--- a/_includes/snippets_library/ComputeCanada_Graham_slurm/scheduler/terminate-multiple-jobs.snip
+++ b/_includes/snippets_library/ComputeCanada_Graham_slurm/scheduler/terminate-multiple-jobs.snip
@@ -12,8 +12,8 @@
 > >
 > > ```
 > > {{ site.remote.prompt }} {{ site.sched.submit.name }} {% if site.sched.submit.options != '' %}{{ site.sched.submit.options }} {% endif %}example-job.sh
-> > {{ site.remote.prompt }} {{ site.sched.submit.name }} {% if site.sched.submit.options != '' %} {{ site.sched.submit.options }}{% endif %}example-job.sh
-> > {{ site.remote.prompt }} {{ site.sched.submit.name }} {% if site.sched.submit.options != '' %} {{ site.sched.submit.options }}{% endif %}example-job.sh
+> > {{ site.remote.prompt }} {{ site.sched.submit.name }} {% if site.sched.submit.options != '' %}{{ site.sched.submit.options }} {% endif %}example-job.sh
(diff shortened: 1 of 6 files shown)


_includes/snippets_library/ComputeCanada_Graham_slurm/scheduler/terminate-multiple-jobs.snip
```diff
@@ -11,9 +11,9 @@
 > > First, submit a trio of jobs:
 > >
 > > ```
-> > {{ site.remote.prompt }} {{ site.sched.submit.name }} {{ site.sched.submit.options }} example-job.sh
-> > {{ site.remote.prompt }} {{ site.sched.submit.name }} {{ site.sched.submit.options }} example-job.sh
-> > {{ site.remote.prompt }} {{ site.sched.submit.name }} {{ site.sched.submit.options }} example-job.sh
+> > {{ site.remote.prompt }} {{ site.sched.submit.name }} {% if site.sched.submit.options != '' %} {{ site.sched.submit.options }}{% endif %}example-job.sh
```
ocaisa, line 14:
```suggestion
> > {{ site.remote.prompt }} {{ site.sched.submit.name }} {% if site.sched.submit.options != '' %}{{ site.sched.submit.options }} {% endif %}example-job.sh
```
Shouldn't the space be like this?

ocaisa:
I'm not sure about this change. I know it works but is not easy to pick apart for someone who doesn't know Jekyll. If the cost of simplicity is an extra space, I think maybe that's ok.

There is one more option:
```
{{ site.sched.submit.name }}{{ site.sched.submit.options }}
```
(remove the space), and add a note in the config file that `site.sched.submit.options` must begin with a space. I don't like that too much either though...

mikerenfro:
> (remove the space), and add a note in the config file that `site.sched.submit.options` must begin with a space. I don't like that too much either though...

None of the options so far are great. I can live with the extra spaces in the final output. Other options include:

- replacing the hard-coded space character with a more explicit/wordy `append` filter inside the scheduler options tag
- smashing the submit command and the scheduler options together, and using `append` and `prepend` filters in the tags
- making our own local Jekyll filter to factor out the `append` and `prepend` commands into a higher-level function, but I'm not sure if that would work directly on GitHub Pages.

tkphd (approved):
The extra template clutter is unfortunate, but extra spaces are even worse. Thanks for addressing this detail @mikerenfro!

PR #360, opened 2021-05-28
Capitalize Environment Modules throughout. Other uses of module can p…

qiwang7:
…robably be lowercase.

See https://github.com/carpentries-incubator/hpc-intro/issues/297

### DIFF
--- a/_episodes/14-modules.md
+++ b/_episodes/14-modules.md
@@ -47,22 +47,22 @@ to work.
 
 ## Environment Modules
 
-Environment modules are the solution to these problems. A *module* is a
+Environment Modules are the solution to these problems. A *Module* is a
 self-contained description of a software package &mdash; it contains the
 settings required to run a software package and, usually, encodes required
 dependencies on other software packages.
 
-There are a number of different environment module implementations commonly
+There are a number of different Environment Module implementations commonly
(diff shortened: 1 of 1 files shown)


tkphd:
Thanks for the contribution, @qiwang7! It's good to have lots of eyes on our lesson materials to make sure we're getting things right.

In this case, my understanding is that we are using "environment modules" in the generic sense, rather than speaking explicitly about the Environment Modules software framework. This is supported by one of the first lines you patched:

> There are a number of different environment module implementations commonly used on HPC systems: the two most common are *TCL modules* and *Lmod*.

I therefore consider the repository more-correct as it already is. If other community members are strongly in favor of this change we can merge it, but at the moment, I am not inclined to do so.

Thanks again for the pull request! Please do continue to look for errors, imperfections, and areas that could use clarification.

mikerenfro (commented):
Agreeing with Trevor. Though I may have had a note to do this sort of capitalization, the way this lesson is written, every use of the word `module` or `environment module` refers to a generic implementation.

Using `TCL modules` as a synonym for the [Environment Modules](http://modules.sourceforge.net) project means that we don't confuse anyone by using both the uncapitalized and capitalized versions of the words.

tkphd:
Thanks again for the suggestion, @qiwang7. I'm closing this PR in favor of the more general grammar we use in the lesson materials.

Issue #116, opened 2019-11-07
key points of first lesson


psteinb:
Check that the key ides are complete and correct for the [first lesson](https://github.com/hpc-carpentry/hpc-intro/blob/gh-pages/_episodes/11-hpc-intro.md) with reference to [lesson-outline.md](lesson-outline.md).

Use to [cdh](https://cdh.carpentries.org/developing-content.html#supporting-elements) as a guideline on how to write lesson objectives.

tkphd:
From The Carpentries' *Curriculum Development Handbook*:

> ## 5.2.2 Key points
>
> While learning objectives communicate to learners the skills they should develop by working through each episode, key points summarize the main pieces of knowledge that learners should remember after completing the episode. These should also be limited to ~5-7 items. For the episode we’ve been considering in the Data Carpentry lesson Introduction to the Command Line for Genomics, the key points are:
>
> * “The shell gives you the ability to work more efficiently by using keyboard commands rather than a GUI.”
> * “Useful commands for navigating your file system include: ls, pwd, and cd.”
> * “Most commands take options (flags) which begin with a -.”
> * “Tab completion can reduce errors from mistyping and make work more efficient in the shell.”
>
> The Carpentries lesson template automatically creates a reference page that includes all of the key points for each episode in the lesson. Key points should be specific enough that learners are able to use this reference page as review.

tkphd:
The current key points are:

> - High Performance Computing (HPC) typically involves connecting to very large computing systems
  elsewhere in the world.
> - These other systems can be used to do work that would either be impossible or much slower on
  smaller systems.
> - The standard method of interacting with such systems is via a command line interface called
  Bash.

PR #365, opened 2021-07-26
Removed mention of bash in Episode 11

cbohn4:
The lesson's key points are correct and match the lesson's content. I believe bash should be removed from the key points since it was not used in the specific episode. For a brief introduction, it would be best to keep it as a simple difference between CLI and GUI until bash is needed in later episodes' examples when running different jobs.

For the key points review in #116


tkphd (approved):
Good catch! LGTM. Approved, pending CI checks.

annajiat (commented):
agreed

PR #372, opened 2022-01-06
Refactor parallelization section

bkmgit:
Adding a reference. Would like to trim material and have participants download the code rather than type it in.


mikerenfro (approved):
Looks good to me. I've been on other tasks for months now, so hadn't seen the hpc-parallel lesson at all. Quite nice. Any further notes from @tkphd @reid-a @psteinb @ocaisa @annajiat ?

PR #373, opened 2022-01-12
Transcribe HPC Shell material

### LINKED ISSUE #333: introduce SSH keys
In our lesson, we use SSH with passwords, which is not the best practice and, on many systems, is not allowed. Connection with a public key, guarded with a strong passphrase, is greatly preferred.
HPC Shell goes in depth on SSH keys. We should introduce it in the introduction. The logical place for this is after connecting the first time, using a passphrase: at that point, `.ssh` on the remote should exist, and the machinery is in place to `rsync` a public key. Discussion does not need to be deep, we simply require a stronger-than-default set of flags supported by all reasonably modern machines.

How's this:

```bash
ssh-keygen -a 100 -b 4096 -t rsa
```

* `-a 100`: run the key derivation function (KDF) 100 times. Default is 16.
* `-b 4096`: generate a 4096-bit key. Default for RSA (the default cipher) is 2048.
* `-t rsa`: use the RSA cryptosystem. This is probably the default.

tkphd:
This PR closes #333 and closes [hpc-shell/#18](https://github.com/hpc-carpentry/hpc-shell/issues/18).


reid-a:
On the first pass, this looks like a solid start.

A few minor points:

- On many systems, especially all modern Linux systems, an SSH agent is started as part of the default desktop environment. The setup does this right, having users check for the existence of an agent before starting one, but the Ep. 12 instructions just say to create a key-pair and then go ahead and launch the agent, which may clobber an existing agent's 'SSH_*` environment variables. 
- Transferring the public key to the remote system seems to be treated as a possible eventuality -- on CTCMS, of course, this is a requirement for access, and any other public-key-based system will have some non-optional process whereby the public key is transported to the remote system. I think this should be punched up, but I confess to not knowing what standard practice is for non-CTCMS systems.
-  `ssh` command arguments are not explained. Not sure how much detail is really necessary, Unix shell graduates will know about command arguments, but the connection between the `-t 8h` in `ssh-add -t 8h` and the eight-hour lifetime of the key is implicit, rather than explicit, in Ep. 12. Here again the setup does a better job, but uses `1d` instead of `8h` (reflecting the prior nature of the setup, maybe?).

tkphd:
Per the [Coordination Meeting](https://codimd.carpentries.org/RkBKQuHsSN6S9Z3DqV6vBQ), the SSH keys material should be moved into the Lesson proper.

_episodes/12-cluster.md
```diff
@@ -101,7 +141,24 @@ supplied by the instructors. You may be asked for your password. Watch out: the
 characters you type after the password prompt are not displayed on the screen.
 Normal output will resume once you press `Enter`.
 
-## Where Are We?
+You may have noticed that the prompt changed when you logged into the remote
+system using the terminal (if you logged in using PuTTY this will not apply
+because it does not offer a local terminal). This change is important because
+it makes it clear on which system the commands you type will be run when you
```
tobyhodges, line 147:
```suggestion
it can help you distinguish on which system the commands you type will be run when you
```

because clarity is in the eye of the beholder

_episodes/13-scheduler.md
```diff
@@ -106,18 +106,29 @@ command (or series of commands) that you want to run on the cluster is called a
 *job*, and the process of using a scheduler to run the job is called *batch job
 submission*.
 
-In this case, the job we want to run is just a shell script. Let's create a
-demo shell script to run as a test. The landing pad will have a number of
-terminal-based text editors installed. Use whichever you prefer. Unsure? `nano`
-is a pretty good, basic choice.
+In this case, the job we want to run is just a shell script -- essentially a
```
tobyhodges, line 109:
```suggestion
In this case, the job we want to run is a shell script -- essentially a
```

remove dismissive language around shell scripting, which some learners may have only relatively recently learned about/how to work with.

_episodes/13-scheduler.md
```diff
@@ -106,18 +106,29 @@ command (or series of commands) that you want to run on the cluster is called a
...
+text file containing a list of UNIX commands to be executed in a sequential
+manner. Our shell script will have three parts:
+
+* On the very first line, add `{{ site.remote.bash_shebang }}`. The `#!`
+  (pronounced "hash-bang" or "shebang") tells the computer what program is
+  meant to process the contents of this file. In this case, we are telling it
+  that the commands that follow are written for the command-line shell (what
+  we've been doing everything in so far). If we wanted our script to be run
+  with something else -- Python, for example -- we could change thios line to
```
tobyhodges, line 118:
```suggestion
  with something else -- Python, for example -- we could change this line to
```

_episodes/13-scheduler.md
```diff
@@ -126,6 +137,103 @@ hostname
...
+```
+{: .error}
+
+Strangely enough, Bash can't find our script. As it turns out, Bash will only
+look in certain directories for scripts to run. To run anything else, we need
+to tell Bash exactly where to look. To run a script that we wrote ourselves, we
+need to tell Bash where it is explicitly, so it doesn't go searching. We could
+do this one of two ways: either with the absolute path 
+`{{ site.workshop_host_homedir }}/yourUserName/example-job.sh` (equivalently,
+`~/example-job.sh`), or with the relative path `./example-job.sh`.
```
tobyhodges, line 158: worth explaining why the `./` is required even when the script is located in the current working directory, as it is here?

_episodes/13-scheduler.md
```diff
@@ -126,6 +137,103 @@ hostname
...
+{: .language-bash}
+
+```
+...
+-rw-rw-r-- 1 yourUsername yourGroup  40 Jan 16 19:41 example-job.sh
+...
+```
+{: .output}
+
+That's a huge amount of output: a full listing of everything in the directory.
```
tobyhodges, line 185: The command above only listed files with the `.sh` extension, and this sentence/paragraph should to be modified to reflect this.

tkphd:
Add a blurb about strong passphrases for the SSH key:
- Long phrase (32 char+) with punctuation. Street addresses work well for this.
- Use a password manager to generate & store a high-entropy string.
- ...

tkphd:
Make sure we don't clobber existing keys. Also, don't reuse an existing one, as we don't know how strong it is.

_episodes/13-scheduler.md
```diff
@@ -126,6 +137,103 @@ hostname
...
+3. **Owner:** This is the username of the user who owns the file. Their
+   permissions are indicated in the first permissions field.
+4. **Group:** This is the user group of the user who owns the file. Members of
+   this user group have permissions indicated in the second permissions field.
+5. **Size of item:** This is the number of bytes in a file, or the number of
+   [filesystem blocks](https://en.wikipedia.org/wiki/Block_(data_storage))
+   occupied by the contents of a folder. (We can use the `-h` option here to
+   get a human-readable file size in megabytes, gigabytes, etc.)
+6. **Time last modified:** This is the last time the file was modified.
+7. **Filename:** This is the filename.
```
tobyhodges, line 220: I like this explanation! I think there are also some good diagrams around to represent `ls-l` output format - perhaps one could be reused here?

reid-a (approved):
This is broadly right, and has reached diminishing returns in the PR process. IMO OK to merge, we can clean up details via other issues and/or PRs.

tobyhodges (approved):

_episodes/13-scheduler.md
tkphd: :+1:

_episodes/13-scheduler.md
tkphd: I've seen a few enumerations of columns, like this one, but can't recall a good diagram or image doing it justice. If you come across one, please let me know, and if it's Creative Commons licensed, I'll include or hotlink it!

PR #377, opened 2022-02-01
Jargon-buster presentation

reid-a:


bkmgit:
This is nice. Would suggest not using PDF as default, but instead use HTML slide software such as:
- https://github.com/remarkjs/remark
- https://revealjs.com/
- https://www.w3schools.com/howto/howto_js_slideshow.asp

reid-a:
The benefit of PDF and the `pdfpc` tool in particular are that it allows for a "presentation console", where speaker notes are available to the presenter. This helps to ensure that all of the relevant points actually get mentioned to learners, without requiring a large amount of text on the slides themselves.

This is high-value, I think, although we have already had some portability issues, newer `pdfpc` implementations apparently don't use the same notes file format as the one used to compose this initially.

If there is an HTML slide tool that can show notes to the presenter, that would be a good solution.

Technically, it's easy to add new targets to the makefile, of course.

tkphd (approved):
LGTM!

Note that the repo currently contains both the original Inkscape-to-PNG images *and* resized versions. Since the slides are now HTML (not PDF), direct use of the SVG would be preferable., but that's not necessarily a gating item.

Issue #362, opened 2021-06-18
Simplifying the parallel application example


ocaisa:
Based on our discussion yesterday, I created a simplified version of an mpi4py example that exactly reproduces Amdahls law. You could use it as a black box, or explain the code.
```
#!/usr/bin/env python
"""
Amdahl law illustrator (with fake work)
"""

from mpi4py import MPI
import sys
import time

def do_work(work_time=30., parallel_proportion=0.5):
    # How many MPI ranks (cores) are we?
    size = MPI.COMM_WORLD.Get_size()
    # Who am I in that set of ranks?
    rank = MPI.COMM_WORLD.Get_rank()
    # Where am I running?
    name = MPI.Get_processor_name()

    # use Amdahls law to calculate the expected speedup for a given workload
    speed_up = 1. / ((1. - parallel_proportion) + parallel_proportion / size)

    # Set the sleep time (which is used to fake the amount of work)
    sleep_time = work_time / speed_up
    total_sleep = sleep_time * size

    if rank == 0:
        sys.stdout.write(
            "Processors will do %s seconds of 'work', which should take %s seconds on %s cores with %s parallel proportion of the workload.\n"
            % (work_time, sleep_time, size, parallel_proportion))

    time.sleep(sleep_time)

    sys.stdout.write(
        "Hello, World! I am process %d of %d on %s and I did 'work' for %s seconds.\n"
        % (rank, size, name, sleep_time))

do_work()
```

This would give output like
```
[ocaisa@node1 ~]$ time mpirun --oversubscribe -n 2 python simple_mpi.py
Processors will do 30.0 seconds of 'work', which should take 22.5 seconds on 2 cores with 0.5 parallel proportion of the workload.
Hello, World! I am process 0 of 2 on node1.int.isc21.learnhpc.eu and I did 'work' for 22.5 seconds.
Hello, World! I am process 1 of 2 on node1.int.isc21.learnhpc.eu and I did 'work' for 22.5 seconds.

real    0m23.859s
user    0m0.401s
sys     0m0.448s
```
so would also exhibit that there is other overhead to consider.  You can oversubscribe this to your hearts content since it is not actually doing any calculations.

ocaisa:
Hmm, I think I can do better here, I can set the actual amount of serial time for the root (default is `0.5 * work_time`), and distribute a trivial parallel sleep time (default would be `(0.5 * work_time)/n_proc`).

bkmgit:
[...]

PR #378, opened 2022-02-03
Add a trivial Amdahl Law reproducer

ocaisa:
Replaces #362


tkphd:
The naming (`amdahl_v1.0`) suggests a submodule or external dependency to me. I knoiw I asked for it here, but @ocaisa are you sure you want to include your code in this repo, rather than setting up a separate, stand-alone Amdahl's Daemon repository? That would give you the ability to tag & release, with documentation & reference plots, and a 'share' folder (or similar) for the Modulefiles. Curious to understand your thinking / philosophy for small, could-stand-alone projects.

(My philosophy is: they who die with the most repos, wins.)

ocaisa:
I did that at https://github.com/ocaisa/amdahl now. I'll check it actually works out-of-the-box tomorrow.

PR #380, opened 2022-02-09
Replacing jargon presentation pngs with svgs where possible.

reid-a:
The workflow picture has bitmap overlays for the arrows
and shading, so there is no SVG equivalent.  Ditto for
the photo of the data center.  This commit includes replacements
for some currently-unused images, specifically the laptop
and the mainframe, to allow for their possible later use.


reid-a:
Noteworthy that, in my local testing, the image captions are not appearing in either the presentation or display modes. Possibly a quirk of my environment?

files/jargon.html
```diff
@@ -49,7 +49,7 @@
 # Your Personal Computer
 
 .center[
-![Standalone computers (banana for scale)](/fig/jargon_standalone.png)
+![Standalone computers (banana for scale)](/fig/jargon_standalone.svg)
```
tkphd, line 52: Figure is too large for the slide.

tkphd:
The stuff in square-brackets for an image link is alt-text, not a caption. It shows up if you inspect the image elements.
For captions, add text on the line immediately following the image declaration.

PR #386, opened 2022-02-23
adding preliminary GPU lesson

stoltzstrop:
This is a preliminary GPU lesson added to target users of Cirrus. The lesson follows the same example given in the MPI parallel lesson of estimating pi with a similar workflow. The example is rewritten in PyCUDA and stepped through in detail. 

Five files are added: 

- a new episode for the lesson 
- two figures which feature in the lesson
- two files containing the PyCUDA example


aturner-epcc:
@stoltzstrop Thanks for contributing this material, I think it looks like the basis of a useful optional addition to the lesson. However, as it will not be part of the core material taught in most workshops, I think this material should be in the `_extras` directory rather than the `_episodes` directory. Can you update the PR to move it.

I also think the job submission script is specific to a particular HPC system (Cirrus at EPCC in this case) so this should be noted in the material. It may be that this will end up being a customisable snippet like in the main episodes but I think that is a little more complicated for an optional episode which may not have a snippet for every system defined for the lesson.

tkphd:
Hi @stoltzstrop, while there are some details to iron out, I think this is a great contribution! Thanks a ton(ne)!

@aturner-epcc this is clearly a Lesson, not an Extra (where we hide Instructor notes, etc.). I feel we should add it as proposed, then rearrange the lesson to achieve a better fit. As it stands, since this lesson has not been added to the top-level `_config.yml`'s [list of lessons](https://github.com/carpentries-incubator/hpc-intro/blob/22fffbcf0edfebc6d61b5c46df2c69f80bd6c722/_config.yml#L65), it will not appear on a "stock" build of the site. Once we integrate it with the site templating & snippet library, as appropriate, we can add it to the top-level config so it builds for everyone. I've made edits to this effect.

Note that an upcoming refactoring will likely replace the pi calculator in this lesson with @ocaisa's [Amdahl code](https://github.com/ocaisa/amdahl), with the pi materials migrated to hpc-parallel-novice for deeper discussion of parallel principles. This material is likely to migrate at that point as well.

bkmgit:
@stoltzstrop Could you make this pull request to https://github.com/hpc-carpentry/hpc-novice which will focus on embarrassingly parallel programs, using calculating pi as an example.  This lesson is being refactored to primarily focus on using a scheduler and submitting a job.

stoltzstrop:
Done, will remove this pull request

PR #393, opened 2022-03-31
Add site.url and site.baseurl to some images

unode:
When rendering the page in a platform other than GitHub, the images were not being shown.

This is something that has been solved in more recent versions of the carpentry template using `{{ relative_root_path }}` but this functionality isn't available in the current template.


tkphd (approved):
LGTM! Thanks for the fix. Looks like it's time to re-sync the Carpentries template!

PR #405, opened 2022-06-07
Fix emphasis of single characters

tkphd:
Apparently, Markdown parsers do not properly interpret __strong__ emphasis in
the middle of a word. This commit addresses this by replacing `__` with `**`,
per the recommended strategy:
<https://www.markdownguide.org/basic-syntax#bold-best-practices>


tkphd:
Thanks -- the build error is fixed.

PR #407, opened 2022-06-22
Amdahl code

tkphd:
This PR replaces the dummy `hpc-intro-data.tar.gz` with a tarballed version of @ocaisa's [amdahl] program, `hpc-intro-code.tar.gz`, and substitutes that black-box program in place of the pi calculator. This is in keeping with our goal of reducing the "programming" aspect of hpc-intro to the bare minimum, focusing instead on the cluster and practicing job submission.

Note that, whereas the upstream [amdahl] package can be installed as a module, the version archived here is a stand-alone Python file named `amdahl`, with defaults of 30 seconds of work and a parallel proportion of 85%.

@reid-a and I are planning to teach this version on Friday, 24 June (two days hence).

[amdahl]: https://github.com/ocaisa/amdahl


ocaisa:
I'm a bit surprised by the tarball, in https://github.com/carpentries-incubator/hpc-intro/pull/378 you argued against going that route (for good reasons) so I wonder why to do that now?

tkphd:
Excellent point @ocaisa. The primary answer is "expediency," as Andrew & I want to teach this tomorrow and do not have Modules on the cluster in question. (Its nascent replacement does!)

In #378 you checked in a tarball with a version number, suggesting a desire to iterate on the code. My view is that a tarball is for releases, not development code, which is why I suggested spinning it out into a separate repository.

We also have a tarball for Ep. 15, which the existing lesson uses to work through `scp` and `tar`, but the contents are never actually used. Substituting a new tarball allowed me to keep the bones of Ep. 15, but convert it to do something of actual use, i.e., downloading locally, uploading to the cluster, extracting, and running `amdahl`.

I also made changes to the `amdahl[.py]` script before tarballing. Some of this is useful and could go upstream, some not. Specifically, 

1. The upstream [`amdahl.py`](https://github.com/ocaisa/amdahl/blob/main/amdahl/amdahl.py) can't be run stand-alone; I added an `if __name__ == "__main__"` clause to make it work.
2. I prepended a `#!/usr/bin/env python3` and made it executable.
3. I dropped the `.py` extension to enhance the mystique and make it more of a black box.

ocaisa:
The upstream package can be installed with `pip3 install --user amdahl` and already comes with the `amdahl` executable as part of that installation, but we can talk about it after your lesson

tkphd:
@ocaisa is that something learners would each do? If so, that would mean explaining pip, which I sure do not want to do. Otherwise, `amdahl` works because pip sets an alias , or... ?

reid-a (approved):
Looks good, OK to merge.

ocaisa:
I think this would have been better off in a workshop specific fork rather than merging it here without further discussion.

reid-a:
I was about to open a secondary issue about the logistics of the Amdahl code. I interpreted the scope of this PR narrowly, primarily about getting the Amdahl code into the lesson and the pi example out.

I confess I was also partially motivated by our upcoming workshop deadline, which might not have been altogether proper.

tkphd:
Yeah... sorry @reid-a but I'm with @ocaisa on this. There was a discussion going on about the Amdahl code, Alan had not yet formally reviewed any of the rest of it, and for tomorrow's purposes, this code has already been merged into the [workshop fork](https://github.com/hpc-workshops/ctcms-hpc-intro), so there was no urgency. I suggest reverting the merge-commit and re-opening this PR for further discussion, since checking in the Amdahl tarball is part of this branch.

reid-a:
OK, new PR is #410. Apologies all around.

ocaisa:
@reid-a @tkphd Thanks a lot for reacting on this so quickly

tkphd:
@ocaisa we try hard to do what's right :-)

Now.... it looks like re-opening a PR is not possible, and reverting a revert is messy, but I'm working on it... :P

Issue #409, opened 2022-06-23
Amdahl code deployment strategy


reid-a:
In PR #407 the new Amdahl code was brought in to this lesson, as a black-box code that shows desirable performance characteristics and is a useful demonstration of the benefits and limits of parallelized code.

The actual deployment strategy used is that of download a tarball of the amdahl code, specifically broken out as part of the workshop, but the Amdahl code actually does have its own repo, and is installable via 'pip' or from git, or by other methods more consistent with HPC best practice.

The tar-ball scheme was partially an expediency to accommodate an upcoming workshop, but this is worth revisiting with a bit more patience with a view to a better long-term soluiton.

PR #411, opened 2022-06-23
Amdahl code, redux

tkphd:
This PR re-opens #407, which was merged prematurely, and relates to #409. Original summary and quoted comments follow.

---

This PR replaces the dummy `hpc-intro-data.tar.gz` with a tarballed version of ~~@ocaisa's~~ our [amdahl] program, `hpc-intro-code.tar.gz`, and substitutes that black-box program in place of the pi calculator. This is in keeping with our goal of reducing the "programming" aspect of hpc-intro to the bare minimum, focusing instead on the cluster and practicing job submission.

Note that, whereas the upstream [amdahl] package can be installed as a module, the version archived here is a stand-alone Python file named `amdahl`, with defaults of 30 seconds of work and a parallel proportion of 85%.

@reid-a and I are planning to teach this version on Friday, 24 June (~~two days hence~~ tomorrow).

---

@ocaisa [wrote](https://github.com/carpentries-incubator/hpc-intro/pull/407#issuecomment-1164064910),

> I'm a bit surprised by the tarball, in #378 you argued against going that route (for good reasons) so I wonder why to do that now?

@tkphd [responded](https://github.com/carpentries-incubator/hpc-intro/pull/407#issuecomment-1164749623),

> Excellent point @ocaisa. The primary answer is "expediency," as Andrew & I want to teach this tomorrow and do not have Modules on the cluster in question. (Its nascent replacement does!)
>
> In #378 you checked in a tarball with a version number, suggesting a desire to iterate on the code. My view is that a tarball is for releases, not development code, which is why I suggested spinning it out into a separate repository.
> 
> We also have a tarball for Ep. 15, which the existing lesson uses to work through `scp` and `tar`, but the contents are never actually used. Substituting a new tarball allowed me to keep the bones of Ep. 15, but convert it to do something of actual use, i.e., downloading locally, uploading to the cluster, extracting, and running `amdahl`.
>
> I also made changes to the `amdahl[.py]` script before tarballing. Some of this is useful and could go upstream, some not. Specifically, 
>
> 1. The upstream [`amdahl.py`](https://github.com/ocaisa/amdahl/blob/main/amdahl/amdahl.py) can't be run stand-alone; I added an `if __name__ == "__main__"` clause to make it work.
> 2. I prepended a `#!/usr/bin/env python3` and made it executable.
> 3. I dropped the `.py` extension to enhance the mystique and make it more of a black box.

@ocaisa [commented](https://github.com/carpentries-incubator/hpc-intro/pull/407#issuecomment-1164771939),

> The upstream package can be installed with `pip3 install --user amdahl` and already comes with the `amdahl` executable as part of that installation, but we can talk about it after your lesson

@tkphd [asked](https://github.com/carpentries-incubator/hpc-intro/pull/407#issuecomment-1164776194),

> @ocaisa is that something learners would each do? If so, that would mean explaining pip, which I sure do not want to do. Otherwise, `amdahl` works because pip sets an alias , or... ?

_Edited to update Amdahl repo URL_

[amdahl]: https://github.com/hpc-carpentry/amdahl


ocaisa:
For me there are a few assumptions going on in general when delivering this course, one of those that you have knowledge/control over the environment of your users. The  [amdahl](https://github.com/ocaisa/amdahl) package is released in the way it is because it can work in a number of ways: if you have modules you can expose it that way (including encoding it's `mpi4py` dependency), if you don't you can still install it via pip (which also checks/installs the required dependency). Just shipping a tarball has a good chance of not working because there is the assumption that `mpi4py` is already available in the users environment.

Ultimately this sounds like something that belongs in a snippet as there a selection of available ways to make the code available which are all perfectly valid:

* Make it available as a module
* Install it via pip
* Download the source itself and ensure you have the dep satisfied somehow before running it (may need some minor tweaking to work out-of-the-box as per @tkphd's previous comment above...but I would not drop the `.py` extension, instead change the command to execute the code)

tkphd:
Thanks for the thoughtful feedback @ocaisa. My rationalization for the tarball is that Ep. 15 (transferring files) depends on a tarball, and previously, it was repackaged data from a Python lesson that is unrelated and is never used after that episode. This meant a lot of the transfer commands were boilerplate with no concrete filenames or paths, which is mystifying and seems counterproductive. Overhauling Ep. 15 with an `amdahl` tarball feels right, because it makes the concepts of transferring more concrete and gives them _purpose_.

You are right, of course, that `mpi4py` may be unavailable, and having the cluster admin install `amdahl` as a module or via `pip` is sensible. Can you update the README/docs in the [amdahl repo]() to outline the recommended workflow for installation, and usage after it's installed?

I can revise the tarball in this PR to contain a driver script that will simply `import amdahl` and run it, so we can still work through the `curl`/`wget`, `tar`, and `scp`/`rsync`  commands in Ep. 15, then run it in Ep. 16.

tkphd:
In today's Coworking Hours, we had a vigorous debate about how to install __amdahl__. The best/obvious option is to `python3 -m pip install amdahl`, which will pull in __mpi4py__ if it's not already installed. Possible breakage modes include
1. The login node is firewalled, and __pip__ can't reach PyPI to get the code.
2. The PyPI version of __mpi4py__ is incompatible with the installed MPI libraries.

Can folks with access try to install __amdahl__ on their login nodes, and check whether `mpirun -np 4 amdahl` successfully runs?

Thanks!

ocaisa:
I read the notes of the [coworking hour where this was discussed](https://codimd.carpentries.org/53bKz2l5Q0SrDIuGVdVaJQ?both) and I still think the easiest is to go with a `pip install` approach.

As mentioned in those notes, it is possible to do an offline `pip install` if you have the tarball available. You can get the latest version of the tarball with the URL https://github.com/hpc-carpentry/amdahl/tarball/main so there would be no need to maintain a versioned tarball in the lesson.
With a tarball on the system (via curl/wget, plus scp if needed), let's just assume that the `$HOME` directory is on a shared FS (usually pretty reasonable assumption). Installation would be:
```
python -m pip install --user <tarball>
```
This doesn't resolve any dependency problem you may have, but the most problematic one for our use case is MPI4PY. This can be installed with
```
env MPICC=/path/to/mpicc python -m pip install --user mpi4py
```
While I haven't tested it, I imagine that `mpi4py` also be replaced by a tarball in this command. The `MPICC` environment variable is only really needed if `mpicc` is _not_ in your environment (or maybe you need a particular MPI C compiler).

tkphd:
Following this month's meetings, I have a few things to change on this branch before it's ready to merge -- mostly, teaching `pip` as the installation method of choice with some caveats for `mpi4py` and offline installation instructions.

tkphd:
Using the Amdahl/MPI4PY tarballs to exercise "transferring files," make sure to cover the cluster configuration firewalls allowing users to "pull" from inside, "push" from outside, or both.

_episodes/15-transferring-files.md
```diff
@@ -278,137 +286,151 @@ The most common archiving command you will use on a (Linux) HPC cluster is
...
+hpc-carpentry-amdahl-46c9b4b/requirements.txt
+hpc-carpentry-amdahl-46c9b4b/setup.py
 ```
 {: .language-bash}
 
-This shows a folder containing another folder, which contains a bunch of files.
-If you've taken The Carpentries' Shell lesson recently, these might look
-familiar. Let's see about that compression, using `du` for "**d**isk
-**u**sage".
+This shows a folder which contains a few files. Let's see about that
```
ocaisa, line 319:
```suggestion
This shows the folder containing the files is actually called
`hpc-carpentry-amdahl-46c9b4b` (not `amdahl` as we might have
expected) and contains a few files. Let's see about that
```

_episodes/15-transferring-files.md
```diff
@@ -278,137 +286,151 @@ The most common archiving command you will use on a (Linux) HPC cluster is
...
+Text files (including Python source code) compress nicely:
+the "tarball" is one-sixth the total size of the raw data!
 
 If you want to reverse the process -- compressing raw data instead of
 extracting it -- set a `c` flag instead of `x`, set the archive filename,
 then provide a directory to compress:
 
 ```
-{{ site.local.prompt }} tar -cvzf compressed_data.tar.gz hpc-intro-data
+{{ site.local.prompt }} tar -cvzf compressed_code.tar.gz amdahl
```
ocaisa, line 411:
```suggestion
{{ site.local.prompt }} tar -cvzf amdahl.tar.gz amdahl
```
Might as well use the "best practice" we indicated already

_episodes/16-parallel.md
```diff
@@ -25,632 +23,430 @@ We now have the tools we need to run a multi-processor job. This is a very
...
-{: .language-python}
+## Install the Amdahl Program
 
-If we run the Python script locally with a command-line parameter, as in
-`python pi-serial.py 1024`, we should see the script print its estimate of
-π:
+With the Amdahl source code on the cluster, we can install it, which will
+provide access to the `amdahl` executable.
+Move into the extracted directory, then use the Package Installer for Python,
+or `pip`, to install it in your ("user") home directory:
```
ocaisa, line 38:
```suggestion
Move into the extracted directory, then use the Package Installer for Python,
or `pip`, to install it in your home directory (using the `--user` flag):
```

_episodes/16-parallel.md
```diff
@@ -25,632 +23,430 @@ We now have the tools we need to run a multi-processor job. This is a very
...
-Most of the calculations required to estimate π are in the
-`inside_circle` function:
+### MPI for Python
 
-1. Generating `n_samples` random values for `x` and `y`.
-1. Calculating `n_samples` values of `radii` from `x` and `y`.
-1. Counting how many values in `radii` are under 1.0.
+The Amdahl code has one dependency: __mpi4py__.
+If it hasn't already been installed on the cluster, `pip` will attempt to
+collect mpi4py from the Internet and install it for you.
```
ocaisa, line 58:
```suggestion
If it hasn't already been installed on the cluster somehow, `pip` will
attempt to collect mpi4py from the Internet and install it for you.
```

_episodes/16-parallel.md
```diff
@@ -25,632 +23,430 @@ We now have the tools we need to run a multi-processor job. This is a very
...
+> If installing Amdahl failed because mpi4py could not be installed,
+> retrieve the tarball from <https://github.com/mpi4py/mpi4py/tarball/master>
+> then `rsync` it to the cluster, extract, and install:
+>
+> ```
+> {{ site.local.prompt }} wget https://github.com/mpi4py/mpi4py/tarball/master -O mpi4py.tar.gz
+> {{ site.local.prompt }} scp amdahl.tar.gz {{ site.remote.user }}@{{ site.remote.login }}:
+> # or
+> {{ site.local.prompt }} rsync -avP amdahl.tar.gz {{ site.remote.user }}@{{ site.remote.login }}:
+> ```
```
ocaisa, line 73:
```suggestion
> {{ site.local.prompt }} scp mpi4py.tar.gz {{ site.remote.user }}@{{ site.remote.login }}:
> # or
> {{ site.local.prompt }} rsync -avP mpi4py.tar.gz {{ site.remote.user }}@{{ site.remote.login }}:
> ```
```

_episodes/16-parallel.md
```diff
@@ -25,632 +23,430 @@ We now have the tools we need to run a multi-processor job. This is a very
...
+> {: .language-bash}
+> ```
+> {{ site.remote.prompt }} mkdir mpi4py
+> {{ site.remote.prompt }} tar -xvzf mpi4py.tar.gz -C amdahl --strip-components=1
```
ocaisa, line 77:
```suggestion
> {{ site.remote.prompt }} tar -xvzf mpi4py.tar.gz -C mpi4py --strip-components=1
```

_episodes/16-parallel.md
```diff
@@ -25,632 +23,430 @@ We now have the tools we need to run a multi-processor job. This is a very
...
+> {{ site.remote.prompt }} cd mpi4py
+> {{ site.remote.prompt }} python3 -m pip install --user .
+> {{ site.remote.prompt }} cd ../amdahl
```
ocaisa, line 80:
```suggestion
> # Now we can go ahead and install amdahl
> {{ site.remote.prompt }} cd ../amdahl
```

_episodes/16-parallel.md
```diff
@@ -25,632 +23,430 @@ We now have the tools we need to run a multi-processor job. This is a very
...
+> `amdahl` program:
+>
+> ```
+> {{ site.remote.prompt }} which amdahl
+> ```
+> {: .language-bash}
+>
+> If the command returns no output, displaying a new prompt, it means the file
+> `amdahl` has not been found. You must update the environment variable named
+> `PATH` to include the missing folder.
```
ocaisa, line 107:
```suggestion
> `PATH` to include the missing folder that `pip` mentioned.
```

_episodes/16-parallel.md
```diff
@@ -25,632 +23,430 @@ We now have the tools we need to run a multi-processor job. This is a very
...
+{: .discussion}
 
-A simple way to measure the calculation time is to use Python's `datetime`
-module to store the computer's current date and time before and after the
-calculations, and calculate the difference between those times.
+`pip` may warn that your user package binaries are not in your PATH.
 
-To add the time measurement to the script, add the following line below the
-`import sys` line:
+> ## If Pip Raises a Warning...
```
ocaisa, line 88:
```suggestion
> ## If `pip` Raises a Warning...
```

_episodes/16-parallel.md
```diff
@@ -25,632 +23,430 @@ We now have the tools we need to run a multi-processor job. This is a very
...
+> {: .language-bash}
+>
+> If the command returns no output, displaying a new prompt, it means the file
+> `amdahl` has not been found. You must update the environment variable named
+> `PATH` to include the missing folder.
+> Run the following command to update your shell configuration file, then log
+> off the cluster and back on again so it takes effect.
+>
+> ```
+> {{ site.remote.prompt }} echo "export PATH=${PATH}:${HOME}/.local/bin" >> ~/.bashrc
```
ocaisa, line 112: Getting this wrong is extremely dangerous and can break their login shell...I can't think of any way out though

_episodes/16-parallel.md
```diff
@@ -25,632 +23,430 @@ We now have the tools we need to run a multi-processor job. This is a very
...
+>
+> The cluster output should be written to a file in the folder you launched the
+> job from.
+>
+> ```
+> {{ site.remote.prompt }} ls -t
+> ```
+> {: .language-bash}
+> ```
+> slurm-347087.out  serial-job.sh  amdahl  README.md  LICENSE.txt
```
ocaisa, line 188: Not sure about the specific use of slurm here... I guess it's ok if we say: for example,...

_episodes/16-parallel.md
```diff
@@ -25,632 +23,430 @@ We now have the tools we need to run a multi-processor job. This is a very
...
-* How long did the job take to run?
+Use `ls` to locate the output file. The `-t` flag sorts in
+reverse-chronological order: newest first. What was the output?
 
-Modify the job script to increase both the number of samples and the amount
-of memory requested (perhaps by a factor of 2, then by a factor of 10),
-and resubmit the job each time.
+> ## Read the Job Output
+>
+> The cluster output should be written to a file in the folder you launched the
```
ocaisa, line 180: Is this true for all schedulers? They should be able to figure out where their output files are from previous content, so perhaps we should jump straight into looking at contents of the output rather than get caught up in scheduler/site specifics

_episodes/17-resources.md
```diff
@@ -72,7 +72,7 @@ information to `less` to make it easier to view (use the left and right arrow
 keys to scroll through fields).
 
 ```
-{{ site.remote.prompt }} {{ site.sched.hist }} {{ site.sched.flag.histdetail }}
+{{ site.remote.prompt }} {{ site.sched.hist }} {{ site.sched.flag.histdetail }} 1965 | less
```
ocaisa, line 75: What's the `1965` for?

ocaisa (changes requested):
I think `hpc-intro-code.tar.gz` is still being shipped with this PR

_episodes/16-parallel.md
ocaisa: Explicitly adding the line via an editor may be the better option

wirawan0: Two things:

1) Beware that some clusters don't use bash as the shell, so it will not work in those esoteric cluster. (Our cluster is one of those, unless the user is savvy enough to switch to bash.)

2) This command leverages two (rather) advanced knowledge as far as the basic Software Carpentry's `shell-novice` lesson is concerned: (a) shell variables; (b) customization using .bashrc. Both are actually covered in the `shell-extra` lesson:

  - https://carpentries-incubator.github.io/shell-extras/07-aliases/index.html
  - https://carpentries-incubator.github.io/shell-extras/08-environment-variables/index.html

Do we need to provide explanation or link for those that may be curious about the syntax, etc?
Learners will need to be warned that it MUST be two `>>` instead of single `>`. It can cause havoc on the user's environment. Another alternative is to edit `~/.bashrc` with nano, though other things could go wrong in this way.

tkphd:
> I think `hpc-intro-code.tar.gz` is still being shipped with this PR

Weird, I thought 8a6f91b832ce6f4e02c37e0e899343d5b7580667 deleted it. Taken care of in decb121, which also consolidates the lingering pi code in a new tarball.

_episodes/15-transferring-files.md
```diff
@@ -59,73 +79,60 @@ mechanism.
...
-```
-{{ site.local.prompt }} scp {{ site.remote.user }}@{{ site.remote.login }}:/path/on/{{ site.remote.name }}/file.txt path/to/local/
+{{ site.local.prompt }} scp local_file {{ site.remote.user }}@{{ site.remote.login }}:remote_path
 ```
 {: .language-bash}
 
 Note that everything after the `:` is relative to our home directory on the
-remote computer. We can leave it at that if we don't care where the file goes.
+remote computer. We can leave it at that if we don't have a more specific
+destination in mind.
```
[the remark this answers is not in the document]
tkphd: Looks good! Minor quibble: `it's` --> `it is` != `its`. So, please revise "(with it's original name)" to "(with its original name)"

_episodes/16-parallel.md
bkmgit: Perhaps rephrase and use something like "on many clusters, this is the folder you launched the job from, but your instructor will provide guidance"

_episodes/16-parallel.md
bkmgit: Editing with nano seems safer and portable.

bkmgit:
Maybe tools like https://www.parallelpython.com/ , https://www.dask.org/ and https://github.com/pgiri/dispy should be mentioned? Not all people will need MPI, and these might provide good alternatives when task farming is the main purpose.

_episodes/17-resources.md
tkphd: That's a dummy JobID

_episodes/16-parallel.md
tkphd: Pip is the proper noun, `pip` is the command (or, more recently, the Python module to invoke). The heading is "correct" as written, but if it causes confusion or increases cognitive loading, I'm open to changing it.

_episodes/16-parallel.md
tkphd: Hrmm. Do we know the naming convention for SGE and Torque/PBS? If so, we could template this; if not, the "for example" idea is solid.

_episodes/16-parallel.md
tkphd: I don't know for certain, but I certainly hope this applies to all schedulers running on Linux clusters. The default behavior is to use the current directory for I/O, unless the submission script specifies something different, which we have not done here. If we make the lesson too generic, it loses value.

tkphd:
> Maybe tools like https://www.parallelpython.com/ , https://www.dask.org/ and https://github.com/pgiri/dispy should be mentioned? Not all people will need MPI, and these might provide good alternatives when task farming is the main purpose.

Out of scope for this PR. Please file a separate issue or, even better, a PR setting MPI in context with these other tools/approaches, bearing in mind that there may be controversy over whether task farming is HPC (vs. HTC).

_episodes/15-transferring-files.md
```diff
@@ -59,73 +79,60 @@ mechanism.
...
+entire directories as well as individual files.
+
+To copy a whole directory, we add the `-r` flag for "**r**ecursive": copy the
 item specified, and every item below it, and every item below those... until it
 reaches the bottom of the directory tree rooted at the folder name you
 provided.
 
 ```
-{{ site.local.prompt }} scp -r some-local-folder {{ site.remote.user }}@{{ site.remote.login }}:target-directory/
+{{ site.local.prompt }} scp -r hpc-carpentry-amdahl-46c9b4b {{ site.remote.user }}@{{ site.remote.login }}:~/
```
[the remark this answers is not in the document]
tkphd: Good catch. Reworked the sequencing so this makes more sense in fac84f2.

_episodes/15-transferring-files.md
```diff
@@ -166,42 +173,40 @@ A trailing slash on the target directory is optional, and has no effect for
...
+>    among other things; implies recursion
+> * `-v` (**v**erbose) to get verbose output to help monitor the transfer
+> * `-P` (partial/progress) to preserve partially transferred files in case
 >   of an interruption and also displays the progress of the transfer.
 >
 > To recursively copy a directory, we can use the same options:
 >
 > ```
-> {{ site.local.prompt }} rsync -avzP path/to/local/dir {{ site.remote.user }}@{{ site.remote.login }}:directory/path/on/{{ site.remote.name }}/
+> {{ site.local.prompt }} rsync -avP hpc-carpentry-amdahl-46c9b4b {{ site.remote.user }}@{{ site.remote.login }}:~/
```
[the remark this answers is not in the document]
tkphd: Yes, good catch! 7b9c78e introduces that as a tid-bit later on, leaving the upload destination blank at first.

_episodes/16-parallel.md
tkphd: Added "for example," snippets for other schedulers is for future-us to resolve.

_episodes/16-parallel.md
tkphd: Agreed, 9b656ad modifies this to use `nano`.

_episodes/17-resources.md
tkphd: Replaced with the previously seen job ID in 470c4c5

tkphd:
I believe the requested changes have been made. Please re-review.

ocaisa (approved):
LGTM

ocaisa:
Many many thanks @tkphd for putting so much effort into this PR!

PR #412, opened 2022-06-30
Removes typo

Kerleyark:
Fixes a minor typo in the beginning of Episode 11 of hpc-intro.


tkphd (approved):
Good find. Thank you for the fix @Kerleyark, and Welcome!

PR #413, opened 2022-07-05
Removes unneeded comma, redundant word, and changes to active voice

Kerleyark:
Modify and Change seem like synonyms, so modify was chosen to be removed.
The comma created a sentence fragment.


tkphd (approved):
Better, thanks!

PR #414, opened 2022-07-13
Episode13 corrections

Kerleyark:
The first commit is adding some clarification to make sure a reader doesn't believe `ls` is meant to read the file contents. 

The second fixes a naming error with the job script. In all the snippets, the job-name given was "new_name". This fixes the issue of the job-name appearing as "hello-world" in the example before that.


tkphd:
Thanks for another contribution, @Kerleyark! The clarifications are appreciated.

I made the change to the script filename (`new_name` to `hello-world`) because learners and inexpert instructors are more likely to just type what's there than to come up with something on the fly, and "new name" is a terrible title. The fix for the discrepancy is to propagate this change through the snippet library (opened as issue #415), not to revert the job name.

If you feel comfortable making that change yourself, please feel free to do so as part of this PR. If not, please revert a38da5b and I'll open a new PR to resolve the new issue.

Kerleyark:
Thank you, @tkphd. I'm glad to help.
I believe I can go through the snippets files and change the variable name from `new_name` to `hello-world`.
I will revert a38da5b shortly and have the snippets fixed.

tkphd (approved):
LGTM, thanks!

PR #416, opened 2022-07-21
Episode17 corrections

Kerleyark:
I decided to split this into two sentences for the sake of making the wording less awkward. It also adds a little clarification.


tkphd (approved):
Much better phrasing, thank you.

PR #421, opened 2022-09-14
Config options for Birmingham Baskerville slurm

jnywong:
New PR to replace https://github.com/carpentries-incubator/hpc-intro/pull/418


reid-a:
It looks like the CI does not yet include the new snippets as one of the things it tests -- that's kind of a chicken-and-egg problem, of course, but it would be a confidence-booster if we knew for a fact that the new snippets actually build. Can the proposers confirm that this is the case?

reid-a:
I have unwound the chicken-and-egg issue by checking it out and building it locally, it does indeed build.

reid-a (approved):
Looks good, OK to merge.

PR #423, opened 2022-09-26
Typo fix for README.md.

reid-a:
This is slightly mysterious -- it *is* an actual error, "script" is meant there, but it's apparently been accepted by the CI for a long time, but now is not being accepted.  "Scrip" is a perfectly legitimate English word, though not what is meant here, so the CI rejection is a bit puzzling also. Maybe the dictionary changed?


ocaisa (approved):

ocaisa:
I've seen this elsewhere as well, probably comes from using the latest version of codespell

PR #431, opened 2023-03-20
HISAT2 website moved

eckhrd:
The HISAT2 website was moved to a new place.


tkphd (approved):
Thanks for the update!

PR #434, opened 2023-04-20
Make sure `sacct` job output is not wrapped in `less`

eckhrd:
I needed to use `less -S`, otherwise less (v. 530) would wrap the text, making it impossible to read.


tkphd (approved):
I like it!

PR #447, opened 2024-02-13
Cdkharris/nodetypes  > qhost

cdkharris:
The template as-written expects us to use one utility both to show a summary of the system (`nodetypes`) and to show details for one node (`qhost -h <node name>`). It sort of looks like this functionality is supposed to be in `nodetypes` but I don't know enough about perl to figure out what's going on there. I've created a new variable so we can refer to `qhost` in the text.


tkphd:
There are some good fixes in this branch, and important (subtle) corrections that we need to incorporate, like the significance of a trailing slash on the target (as written) or the source (as proposed, and which is correct). Thanks for bringing this up @cdkharris, even if it was accidental :grinning:

cdkharris:
It was indeed accidental 😅

PR #453, opened 2024-07-25
Filezilla remote tunneling instructions

cdkharris:
Adds a site snippet to provide instructions for using Filezilla with tunnels. At UCL many users need to connect to the cluster remotely through our SSH bastion without using our VPN, so we provide instructions for using tunnels to do this.

### DIFF
--- a/_episodes/16-transferring-files.md
+++ b/_episodes/16-transferring-files.md
@@ -340,8 +340,8 @@ If your home directory _is_ the destination, you can leave the destination
 field blank, or type `~` -- the shorthand for your home directory -- for
 completeness.
 
-With `scp`, a trailing slash on the target directory is optional, and has
-no effect. It is important for other commands, like `rsync`.
+With `scp`, a trailing slash on the target directory is optional, and has no effect.
+A trailing slash on a source directory is important for other commands, like `rsync`.
 
 > ## A Note on `rsync`
 >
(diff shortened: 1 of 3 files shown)


reid-a:
Thanks for this, this looks useful. It's breaking the CI, it looks like some of the configs don't like it. I'll have a look, but it might be an easier fix at your end?

cdkharris:
Looks like the problem is that I created a new snippet file for our site, but I haven't created the corresponding files for the other sites. I'll add some blank files for these snippets.

Before I add those, is this a useful addition to the main lesson? It's useful for us, but I don't know whether it's useful for other sites.

reid-a:
Honestly not sure. There are two reasons not to do it, firstly it might be specific to your site, and secondly we have a long-standing TODO about moving our repo away from snippets when we migrate it to the Carpentries Workbench.

My own personal opinion is that we should err on the site of inclusiveness and enriched functionality, and that doing this is a small increase in migration work, so I favor it. But I am not the one doing the work, either. So if you're looking for encouragement, consider yourself encouraged, but not much more than that!

cdkharris:
Understood; I will add this change because it's possible that someone else will be interested in using SSH tunnels with Filezilla, but I'll keep anything more site-specific than this in our repo.

reid-a (approved):
Looks good, thanks!

PR #459, opened 2025-01-13
Skeleton for a new cluster 😁

tkphd:
Snippets for a new cluster at NIST.


ocaisa:
The CI failure is coming from another set of snippets, must be an updated version of codespell

tkphd:
Weird. "Fram" is in the top-level codespell ignore list, and it does not throw errors when I `make spellcheck` locally. Maybe it rtuns in a different folder on the CI? Dunno. Removed the offending term, since the library addresses SAGA, not Fram.

reid-a:
Looks good now.

reid-a (approved):

PR #463, opened 2025-04-17
Boulder feedback

tkphd:
Some small changes based on feedback from the NIST Boulder workshop.

### DIFF
--- a/_episodes/14-environment-variables.md
+++ b/_episodes/14-environment-variables.md
@@ -232,12 +232,9 @@ here are the components of `PATH` listed one per line:
 ~~~
 {: .output}
 
-On our computer,
-there are actually three programs called `analyze`
+Consider a computer with three different programs, all called `analyze`,
 in three different directories:
-`/bin/analyze`,
-`/usr/local/bin/analyze`,
-and `/users/vlad/analyze`.
+`/bin/analyze`, `/usr/local/bin/analyze`, and `/users/vlad/analyze`.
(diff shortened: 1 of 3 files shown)


bkmgit:
Any reason for removing Filezilla? For people new to the command line a GUI file transfer environment can be a helpful option.

reid-a:
There are a couple of reasons for getting rid of Filezilla. The most important to me is that it detracts from the focus of the lesson, which is all about the power and utility of the command line. It's kind of a weird one-off that doesn't mesh well with the rest of the material, and so is cognitively burdensome.

There are secondary reasons. By reputation, it apparently used to install spyware even when installed from the official repos. It also isn't compatible with emerging security requirements like MFA -- this isn't a deal-breaker now, but it will be soon. IMO.

I think there are also other graphical file-transfer tools that are better? Many modern feature-ful text editors can apparently do this, including VSCode, and I think Sublime Text and Atom. I know they can all edit remote files, but I'm less sure about file transfers for the last two.

bkmgit:
Being able to view file structure hierarchy is helpful. Cross platform tools are few. One alternative is Midnight Commander:
- https://midnight-commander.org
- https://github.com/adamyg/mcwin32

It probably deserves a place in the lesson, though would be an optional callout - if someone is reading the lesson on their own, or after participating in a workshop, pointers that allow them to get unstuck are particularly valuable.

Koki1610 (approved):

Koki1610 (approved):

PR #483, opened 2026-03-20
Modified the global config to comment out the array lesson.

reid-a:
Can be restored once Issue #482 (templating for this lesson) is resolved

The purpose of this is, it's a simple way to keep the public repo in a functional state, without reverting the merge of PR #481.


ocaisa (approved):
LGTM

PR #492, opened 2026-04-19
Carol ep22 parallel primes

carolbooth2:
fixes to our alernative parallel episode:
need to load OpenMPI to compile
incorrect laptop prompt changed to comet prompt for copying the binaries to a working directory

still to do (can't create an issue):
remove last section [How Much Does Parallel Execution Improve Performance?](http://127.0.0.1:4321/22-parallel.html#how-much-does-parallel-execution-improve-performance) because it's from the Amdahl episode.  Leave it out and send learners to Amdahl episode [How Much Does Parallel Execution Improve Performance?](http://127.0.0.1:4321/17-parallel.html#how-much-does-parallel-execution-improve-performance) if they want to go further with parallel jobs.


tkphd:
I think this PR was made against the wrong upstream repository.

bkmgit:
Some of the material seems useful.  The job arrays are likely a common workflow for many people using an HPC cluster and would led to better utilization of resources as tasks could be scheduled without requiring good communication infrastructure needed for MPI jobs. This is being added in:
- https://github.com/carpentries-incubator/hpc-intro/pull/481
- https://github.com/carpentries-incubator/hpc-intro/pull/489


Did learners find the prime calculation useful?
The code is pulled from https://github.com/NewcastleRSE-Training/HPC_Training_Example_Jobs/blob/main/c
rather than put in the main repository.  Would it be helpful if the repository was structured and had guidelines to enable easy addition and updates of the examples?

For primes calculation, would expect that task based parallelism would be a better approach to avoid load imbalance for the current algorithm, with MPI, use one core as a manager which sends a portion of the work of to worker cores. Once a worker has finished processing a range of numbers, it can send these back and ask for more work.  For an MPI4PY example see:
https://mpi4py.readthedocs.io/en/stable/mpi4py.futures.html

Do many of the target learners know C?  MPI does not have direct support for multiprecision numbers, so while the example is interesting, extending it is non trivial.

carolbooth2:
not intended for this repo. Was intending to update our fork only

PR #493, opened 2026-04-21
Populate glossary for Instructor Notes

TejeshPala:
- resolves https://github.com/carpentries-incubator/hpc-intro/issues/37

PR #495, opened 2026-05-06
Move glossary to correct location and format

ocaisa:
Follow up to #493 to put the glossary in the place where it is found by the heading. Looks like the current best practice is to use glosario...but I just spent 2 hours doing this and we need a fix


instructors/instructor-notes.Rmd
```diff
@@ -1,262 +1,3 @@
 ---
-title: Glossary
-permalink: /guide/
```
ocaisa, line 3: These are hangovers from the old format I think

learners/reference.md
```diff
@@ -44,8 +43,8 @@ instead of showing the transfer rate in real time, the following command fails
 mysteriously.
 
 ```bash
-{{ site.local.prompt }} rsync —progress my_precious_data.txt {{ site.remote.user }}@{{ site.remote.login }}
-rsync: link_stat "/home/{{ site.local.user }}/—progress" failed:
+$ rsync —progress my_precious_data.txt user@host
```
ocaisa, line 46: Getting the templating to work in this file is hard, as the location of the configuration files is no longer the same

PR #500, opened 2026-05-11
Fix: Scheduler options explanation

TejeshPala:


tkphd:
Thank you @TejeshPala, it's a good start. The spacing after commands changes depending on short flags vs. long, eg., `--ntasks=4` vs. `-n 4`.
It might be helpful to write a simple formatting function into `load_config.R` to print the flag-separator-value concatenation, but that would make the config files messier.

Anyway, the change made in this commit has already been made and reverted previously, so merging it will not fix the problem. I recommend against merging.

episodes/13-scheduler.Rmd
```diff
@@ -203,30 +203,49 @@ Fantastic, we've successfully changed the name of our job!
...
+  ranks or processes) should `r config$sched$name` launch?
 
-- `--time <days-hours:minutes:seconds>` or `-t <days-hours:minutes:seconds>`:
+- `--ntasks-per-node=<ntasks>`: How many tasks should be launched on each compute
+  node?
+
+- `--cpus-per-task=<ncpus>` or `-c <ncpus>`: How many CPUs should be allocated
+  to each task/process?
+
+- `--gpus=<count>`: Request GPUs for the job allocation.
```
ocaisa: I don't love this, as whether this is the right way to request  GPUs is system-specific (many sites still use `gres`). I would even be inclined to leave this out to avoid engaging in that discussion.

ocaisa (changes requested):
In principle I'm ok with this change, but I note that the syntax is all Slurm specific. We are not offering another option today, so that's perhaps ok but these will probably need to end up in a snippet when another option like `flux` is introduced.

episodes/13-scheduler.Rmd
TejeshPala: Generalize with the GPUs request as `--gpus` or `--gres` is site-specific use cases.
Sometimes, partition dependent and soon.

ocaisa (approved):
LGTM

ocaisa:
Ah, I see @tkphd has hidden a  review in https://github.com/carpentries-incubator/hpc-intro/pull/500#issuecomment-4430302690 , maybe he wants to comment before we consider merging? I note that Slurm mostly treats ` ` or `=` as equivalent, but then we get into the weeds about which is clearer/better.

ocaisa:
> Thank you @TejeshPala, it's a good start. The spacing after commands changes depending on short flags vs. long, eg., `--ntasks=4` vs. `-n 4`. It might be helpful to write a simple formatting function into `load_config.R` to print the flag-separator-value concatenation, but that would make the config files messier.
> 
> Anyway, the change made in this commit has already been made and reverted previously, so merging it will not fix the problem. I recommend against merging.

One "solution" is to add a comment to the config file saying we always want to use the long form options for clarity. Then we can explicitly add the `=` in the lesson and avoid that it looks like a typo. In `load_config.R` we could then chomp any trailing `=` by default. We could even have `load_config.R` complain if a short form option is used.

PR #501, opened 2026-05-12
fix:

TejeshPala:
- replace analyze placeholder with `bash` application
- use correct env variables explanation and documentation from `set`


ocaisa (approved):
LGTM

PR #518, opened 2026-09-03
Adjust config to use Workbench customization integration

tobyhodges:
Sets the new `workbench_base_customization` parameter and updates the surrounding comment lines to reflect the new way of specifying an overlay customization.


tobyhodges:
Failing build is confusing, since I ran into no such error when I built the branch locally. I will explore further

tobyhodges:
Ah, I was being tricked by caching once again. The episode files were still calling the `load_config.R` script, which is no longer needed, and that was causing some kind of clash. I have now removed those calls, and the `load_config.R` script itself and the build looks healthy.

BTW the instructions in `README.md` are really out-of-date now and I have not attempted to fix that in this PR. But I will open an issue...

froggleston:
I was just about to post this exact solution - good work @tobyhodges !

froggleston (approved):
Looks good!

TejeshPala:
tested and can be merged! 🎉

